PR #17, opened 2023-03-30
Ethercat to ipc

omarhogni:


.idea/cmake.xml
jbbjarnason, line 1: This file can now be removed

libs/ec/inc/public/tfc/cia/402.hpp
```diff
@@ -0,0 +1,127 @@
+#pragma once
+
+#include <boost/sml.hpp>
```
jbbjarnason, line 3: seems that sml is not used, maybe change to 
```C++
#include <cstdint>
#include <string>
```

libs/ec/inc/public/tfc/cia/402.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+namespace tfc::ec::cia_402 {
+
+enum commands_e : uint16_t {
```
jbbjarnason, line 7: we should use strict typing, either `enum struct` or `enum class` so it cannot be implicitly converted to any type that can be implicitly converted to uint16_t

libs/ec/inc/public/tfc/devices/abt/easycat.hpp
```diff
@@ -0,0 +1,85 @@
+#pragma once
+
+#include <bitset>
+#include <cstdio>
+
+#include "ethercat.h"
+#include "tfc/devices/base.hpp"
```
jbbjarnason, line 7: in terms of the folderstructure should this perhaps be `tfc/ec/devices/base.hpp`

libs/ec/inc/public/tfc/devices/abt/easycat.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include "tfc/ipc.hpp"
+
+namespace tfc::ec::devices::abt {
+class easyecat : public base {
+public:
+  explicit easyecat(boost::asio::io_context& ctx) : base(ctx) {}
+  static constexpr auto vendor_id = 0x79a;
+  static constexpr auto product_code = 0xdefede;
+  auto process_data(uint8_t* input, uint8_t* output) noexcept -> void final {
+    if (input == nullptr) {
```
jbbjarnason, line 18: maybe add `|| output == nullptr` so we won't accidentally deref nullptr below

libs/ec/inc/public/tfc/devices/base.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <boost/asio.hpp>
```
jbbjarnason, line 3: I think you can include `boost/asio/io_context.hpp` which should reduce the include size

libs/ec/inc/public/tfc/devices/base.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include "tfc/ecx.hpp"
+
+namespace tfc::ec::devices {
+class base {
+public:
+  explicit base(boost::asio::io_context& ctx) : ctx_(ctx) {}
+  static constexpr auto vendor_id = 0x0;
```
jbbjarnason, line 11: what type is deduced as auto here?

libs/ec/inc/public/tfc/devices/base.hpp
```diff
@@ -0,0 +1,29 @@
...
+  static constexpr auto product_code = 0x0;
+  virtual ~base() = default;
+  // Default behaviour no data processing
+  virtual void process_data(uint8_t*, uint8_t*) noexcept {};
+  // Default behaviour, no setup
+  virtual auto setup(ecx_contextt*, uint16_t) -> int { return 1; };
+
+protected:
+  boost::asio::io_context& ctx_;
```
jbbjarnason, line 20: do we need this reference?, I don't think it is nice to keep references in long living classes, but this should probably be safe.

libs/ec/inc/public/tfc/devices/device.hpp
```diff
@@ -0,0 +1,46 @@
...
+using tfc::ec::devices::schneider::atv320;
+
+template <typename device>
+auto devices_equal(auto vendor_id, auto product_code) {
+  return vendor_id == device::vendor_id && product_code == device::product_code;
+}
+
+auto get(boost::asio::io_context& ctx, auto vendor_id, auto product_code) -> std::unique_ptr<base> {
+  if (devices_equal<atv320>(vendor_id, product_code))
+    return std::make_unique<atv320>(ctx);
```
jbbjarnason, line 31: later we should add devices to mp::list https://github.com/boost-ext/mp and then this function would iterate over defined types

libs/ec/inc/public/tfc/devices/abt/easycat.hpp
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+private:
+  auto set_output(size_t position, bool value) -> void { output_states_.set(position, value); }
+  std::bitset<4> output_states_;
+  std::array<bool, 4> last_bool_value_;
+  std::array<uint8_t, 2> last_analog_value_;
+  std::vector<std::shared_ptr<tfc::ipc::bool_send>> bool_transmitters_;
+  std::vector<std::shared_ptr<tfc::ipc::uint_send>> analog_transmitters_;
+  std::vector<std::shared_ptr<tfc::ipc::bool_recv_cb>> bool_receivers_;
```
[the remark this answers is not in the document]
omarhogni: Issue created

libs/ec/inc/public/tfc/devices/abt/easycat.hpp
```diff
@@ -0,0 +1,85 @@
...
+        });
+      }
+      last_analog_value_[i] = value;
+    }
+    *output = output_states_.to_ulong() & 0x0f;
+  }
+  auto setup(ecx_contextt*, uint16_t slave_index) -> int final {
+    for (size_t i = 0; i < 4; i++) {
+      std::expected<std::shared_ptr<tfc::ipc::bool_send>, std::error_code> ptr =
+          tfc::ipc::bool_send::create(ctx_, fmt::format("easyecat.{}.bool.in.{}", slave_index, i));
```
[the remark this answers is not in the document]
omarhogni: Issue created.

libs/ec/inc/public/tfc/devices/util.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <units/isq/si/electric_current.h>
```
jbbjarnason, line 3: you should not need any include here

maybe add the mapping function to a namespace

libs/ec/inc/public/tfc/devices/base.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <boost/asio.hpp>
+
+#include "tfc/ecx.hpp"
```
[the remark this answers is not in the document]
omarhogni: It is the prefix of all of the V2 functions inside SOEM. my thought was that a function we mask in c++
would be named using the rule.
```cpp
// In c
ecx_readSDO(...);

// c++
ecx::readSDO(...);
```

And that is where the name comes from.

How do you like `soem_interface.hpp`?

libs/ec/inc/public/tfc/ec.hpp
```diff
@@ -0,0 +1,293 @@
...
+   * per context but not per slave. So this class instance
+   * is embeded in the void* userdata. This function is a middle
+   * function to relay the callbacks to the correct slave functions
+   * with data.
+   * @param context
+   * @param slave_index
+   * @return the number of slaves configured
+   */
+  static auto slave_config_callback(ecx_contextt* context, uint16_t slave_index) -> int {
+    auto* instance = static_cast<context_t*>(context->userdata);
```
jbbjarnason, line 255: I would suggest self instead of instance

libs/ec/inc/public/tfc/ec.hpp
```diff
@@ -0,0 +1,293 @@
...
+  ec_eepromSMt eep_SM_;
+  ec_eepromFMMUt eepFMMU_;
+
+  // Timing related variables
+  std::chrono::nanoseconds min_cycle_;
+  std::chrono::nanoseconds max_cycle_;
+  std::chrono::nanoseconds last_cycle_;
+  std::chrono::nanoseconds cycle_sum_;
+  size_t number_of_cycles_;
+  std::array<std::byte, 4096> io_;
```
jbbjarnason, line 291: what is 4096

libs/ec/inc/public/tfc/devices/base.hpp
omarhogni: These are both unsigned 32-bit integers. Changed this to a concrete type for all devices.

CMakeLists.txt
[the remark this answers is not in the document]
omarhogni: This ended up being quite a big rebase. Propably as a result of the github migrate.

libs/ec/inc/public/tfc/devices/abt/easycat.hpp
omarhogni: I could do this check above ( where process_data is called ) but taking the parameter in as a reference is confusing
as this is not a "uint8_t" this is just a pointer to a buffer of memory. 

I would have prefered to use something that verifies the size of the input and does not have the option of being a nullptr. This function signature is something I would like to change.

libs/ec/inc/public/tfc/devices/base.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+namespace tfc::ec::devices {
+class base {
+public:
+  explicit base(boost::asio::io_context& ctx) : ctx_(ctx) {}
+  static constexpr auto vendor_id = 0x0;
+  static constexpr auto product_code = 0x0;
+  virtual ~base() = default;
+  // Default behaviour no data processing
+  virtual void process_data(uint8_t*, uint8_t*) noexcept {};
```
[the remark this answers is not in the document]
omarhogni: default_device is used to deal with devices that we don't have drivers for. I could also throw a runtime error but this was quite useful while most devices I had connected did not have drivers. I could change this s.t. constructing this device logs an error 
```
Device not implemented name: (), vendor id: (), product code: () 
```

thoughts?

libs/ec/inc/public/tfc/devices/base.hpp
omarhogni: The only place these classes reference the ctx is to construct signals and slots in the setup method. I want to change the constructor to accept the slave_index at construction.

Then i can create all slots/signals in the constructor and can remove the reference.

libs/ec/inc/public/tfc/devices/base.hpp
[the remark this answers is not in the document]
jbbjarnason: `soem_interface.hpp` sounds good

libs/ec/inc/public/tfc/devices/base.hpp
omarhogni: I have implemented this change. This also has the added benefit that future device interfaces are less likely to construct signals/slots in the setup method.

libs/ec/inc/public/tfc/devices/util.hpp
omarhogni: Will be used later.

PR #35, opened 2023-04-13
Bring back static analysis

jbbjarnason:


jbbjarnason:
reopen when this will be addressed

PR #45, opened 2023-04-19
dbus asio added to vcpkg

jbbjarnason:


cmake/presets/cfg-vcpkg.json
```diff
@@ -43,10 +43,10 @@
       "name": "cfg-vcpkg-clang-dynamic",
```
jbbjarnason, line 43: this was also broken

PR #71, opened 2023-05-24
Added some documentation and tests

omarhogni:


libs/ipc/inc/public/tfc/ipc/details/dbus_server_iface.hpp
```diff
@@ -248,18 +259,39 @@ class ipc_manager_client {
         std::bind_front(&ipc_manager_client::match_callback, this));
   }
 
+  /**
+   * Register a signal with the ipc_manager service running on dbus
+   * @tparam message_handler the type of the message handler should only accept a std::error_code reference
```
jbbjarnason, line 264: I would recommend using `std::invocable<std::error_code> auto&& handler` instead in the function signature

libs/ipc/inc/public/tfc/ipc/details/dbus_server_iface.hpp
```diff
@@ -248,18 +259,39 @@ class ipc_manager_client {
...
+   * @param name the name of the signal to be registered
+   * @param type  the type enum of the signal to be registered
+   * @param handler  the error handling callback function
+   */
   template <typename message_handler>
   auto register_signal(const std::string& name, type_e type, message_handler&& handler) -> void {
     connection_->async_method_call(handler, ipc_ruler_service_name, ipc_ruler_object_path, ipc_ruler_interface_name,
```
jbbjarnason, line 271: missing std::forward of handler, probably in more cases in this file

libs/ipc/inc/public/tfc/ipc.hpp
```diff
@@ -30,49 +30,45 @@ auto constexpr register_cb(std::string const& ipc_name) {
  * addition with implementing the ipc connection.
  * @tparam type_desc The type description for the slot.
  */
-template <typename type_desc>
+template <typename type_desc, typename manager_client_type>
```
jbbjarnason, line 33: I would recommend populating the default for readability

libs/ipc/inc/public/tfc/ipc.hpp
```diff
@@ -30,49 +30,45 @@ auto constexpr register_cb(std::string const& ipc_name) {
...
+  auto operator=(slot&&) noexcept -> slot& = delete;
+  auto operator=(slot const&) -> slot& = delete;
 
   [[nodiscard]] auto value() const noexcept { return slot_->get(); }  // todo: value() or get()
 
 private:
   static constexpr std::string_view self_name{ slot_tag };
   std::shared_ptr<details::slot_callback<type_desc>> slot_;
-  ipc_ruler::ipc_manager_client client_;
+  manager_client_type& client_;
```
jbbjarnason, line 63: I would prefer sharing ownership of client with smart pointer, as this can result in undefined behaviour.

jbbjarnason (approved):

libs/ipc/inc/public/tfc/ipc.hpp
omarhogni: I don't quite agree, using smart pointers can also result in undefined behavior. If using smart pointers we are enforcing the usage of the component forcing them to create smart pointers.

If the owner of the object wishes to use smart pointers to manage the lifetime of ipc_manager_client when using a reference he can also use a smart pointer. he is just not forced to.

f.e. 
```cpp
#include <iostream>
#include <string>
#include <memory>

struct foo{
  void scream(){
    std::cout << "AHHH" << std::endl;
  }
};

void bar(foo& f){
  f.scream();
}
int main(){
  std::shared_ptr<foo> tee = std::make_shared<foo>();
  bar(*tee);
}
```

PR #77, opened 2023-05-30
dbus confman changes for frontend

jbbjarnason:
- postfix confman paths/interface with `config`
    - use underscore instead of dash


SindriTh (approved):
Looks good

omarhogni (approved):

PR #86, opened 2023-06-08
Make docker toolchain and tighten tests

jbbjarnason:


containers/start-dbus.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+dbus-daemon --system
+
+sed -i 's|deny own=|allow own=|g' /usr/share/dbus-1/system.conf
+sed -i 's|deny send_type="method_call"|allow send_type="method_call"|g' /usr/share/dbus-1/system.conf
+dbus-send --system --type=method_call --dest=org.freedesktop.DBus / org.freedesktop.DBus.ReloadConfig
```
omarhogni, line 6: Check this out, dont need to reload if we set the settings first.

omarhogni: sed can be done in container file

omarhogni (changes requested):
Very good change

containers/start-dbus.sh
jbbjarnason: changed

PR #96, opened 2023-06-14
Fix arm64 install

jbbjarnason:
Fix arm64 install on ubuntu. Have not tried the rpm installer for arm.

Also now throwing runtime error if something goes wrong in creating a signal, like for example if you had run your process with root and than back to normal user. Normal user won't have permission to open socket file if the previous process did not clean up.


libs/ipc/inc/public/tfc/ipc.hpp
omarhogni, line 1: Þessi skrá ekki í þetta PR. std::shared pointerinn þarf ekki að vera constructed instantly má bara vera í constructor. 

Þarf ekki þessa RAII pælingu.

omarhogni (approved):

PR #103, opened 2023-06-16
ATV320 configuration parameters switched to mp-units dimensions

omarhogni:


exes/ec/tests/test_ec_util.cpp
```diff
@@ -62,4 +62,10 @@ auto main(int, char**) -> int {
     expect(exp.has_value() >> fatal);
     expect(exp.value() == test);
   };
+
+  "Test atv320 custom units"_test = []() {
+    static_assert(std::chrono::seconds{ 1 } == tfc::ec::devices::schneider::deciseconds{ 10 });
```
jbbjarnason, line 67: do you wan't to have this here?

omarhogni: For now yes, later I want to create a test just for the atv320.

PR #105, opened 2023-06-19
allow non root processes to connect to root sockets

jbbjarnason:
tried fchmod on the file descriptor from zmq but with no luck(which I would prefer). chmod on the filename works. 
maybe this should be a compile flag but I think it should first become a problem before we make it a flag.


jbbjarnason:
needs more discussion, reopen if needed.

PR #106, opened 2023-06-20
add c++20 coroutines for detail signal and slot

jbbjarnason:
using asio::async_compose for async_send and async_receive for base types of signal and slot

### DIFF
--- a/cmake/tfc_options.cmake
+++ b/cmake/tfc_options.cmake
@@ -58,7 +58,7 @@ if(ENABLE_SANITIZATION)
         -fPIC
         -fno-omit-frame-pointer
         -g )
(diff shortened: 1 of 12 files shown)


cmake/tfc_options.cmake
```diff
@@ -58,7 +58,7 @@ if(ENABLE_SANITIZATION)
         -fPIC
         -fno-omit-frame-pointer
         -g )
-    add_link_options( -lasan -lubsan -ltsan)
```
omarhogni, line 61: Should we not figure out why this needs to be removed?

omarhogni (approved):

PR #110, opened 2023-06-29
propagate json schema type correctly of std::duration and units::quan…

jbbjarnason:


exes/ec/inc/public/tfc/ec/devices/util.hpp
```diff
@@ -50,8 +50,9 @@ template <ecx::index_t idx,
           auto... value_t_construct_params>
 struct glz::meta<tfc::ec::util::setting<idx, name_value, desc, value_t, value_t_construct_params...>> {
   using setting = tfc::ec::util::setting<idx, name_value, desc, value_t, value_t_construct_params...>;
+  static constexpr std::string_view description{ desc };
   static constexpr auto value{
-    glz::object("value", &setting::value, "index", &setting::index, "name", &setting::name_v, "desc", &setting::desc_v)
+    glz::object("value", &setting::value, description)  // todo concat description, index and name
```
SindriTh, line 55: Better,
This might need changing depending on the UI form library used.

SindriTh (commented):
Looks good.

Needs a a bit of work to be fully UI compatible.
That could also be done as a separate PR, its up to you @jbbjarnason

exes/ec/inc/public/tfc/ec/devices/util.hpp
jbbjarnason: will be addressed in future PR, reverted the changes

PR #115, opened 2023-07-04
Mqtt broadcaster that reads signals off of ipc and forwards them to MQTT broker

magni-mar:


exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+
+  asio::io_context& ctx_;
+  ipc_client_type ipc_client_;
+  std::shared_ptr<mqtt_client_type> mqtt_client_;
+  tfc::logger::logger logger_;
+  std::unique_ptr<sdbusplus::asio::connection, std::function<void(sdbusplus::asio::connection*)>> dbus_connection_;
+  std::unique_ptr<sdbusplus::bus::match::match, std::function<void(sdbusplus::bus::match::match*)>> signal_updates_;
+  tfc::confman::config<config> config_;
+
+  std::vector<std::string> banned_signals_;
```
jbbjarnason, line 293: it shouldn't be necessary to make a copy of the config, you can access the config vector from the config member

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+      connect_active_ = true;
+      co_await asio::co_spawn(mqtt_client_->strand(), connect_to_broker(), asio::use_awaitable);
+    }
+
+    // When a new list of banned string arrives the program is restarted
+    config_.value()._banned_topics.observe([this](auto& new_conf, [[maybe_unused]] auto& old_conf) {
+      banned_signals_.clear();
+      for (auto const& con : new_conf) {
+        banned_signals_.push_back(con);
+      }
```
jbbjarnason, line 60:
```suggestion
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+      restart();
+    });
+
+    asio::co_spawn(mqtt_client_->strand(), tfc::base::exit_signals(ctx_), asio::detached);
+
+    banned_signals_ = config_.value()._banned_topics.value();
```
jbbjarnason, line 66:
```suggestion
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+    load_signals();
+  }
+
+  // Connect the mqtt client to the broker. Only a single "thread" can run this at a time. The routine that calls this
```
jbbjarnason, line 70: it is not really a multi thread problem, this routine can only be spawned once.

jbbjarnason: you probably could change the connect flag to a strand to make this safe

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+        co_await asio::async_connect(mqtt_client_->next_layer(), resolved_ip, asio::use_awaitable);
+
+        logger_.trace("Sending Connect request to the MQTT broker");
+
+        // Connect is initialized with clean start set to false, which means that the broker will keep the session alive even
+        // if the client disconnect, and it will keep that session alive for UINT_MAX seconds.
+        auto connect_packet =
+            async_mqtt::v5::connect_packet{ false,
+                                            std::chrono::seconds(100).count(),
+                                            async_mqtt::allocate_buffer("cid1"),
```
jbbjarnason, line 94: could we make a better client id than cid1? What happens if two PCs connect with the same cliend id to one mqtt broker?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+                                            async_mqtt::nullopt,
+                                            async_mqtt::allocate_buffer(mqtt_username_),
+                                            async_mqtt::allocate_buffer(mqtt_password_),
+                                            { async_mqtt::property::session_expiry_interval{ UINT_MAX } } };
```
jbbjarnason, line 98: use C++ https://en.cppreference.com/w/cpp/types/numeric_limits/max instead of C

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+
+        logger_.trace("Waiting for MQTT connection acknowledgement");
+        co_await mqtt_client_->recv(async_mqtt::filter::match, { async_mqtt::control_packet_type::connack },
+                                    asio::use_awaitable);
+
+        connect_active_ = false;
+        co_return;
+      } catch (std::exception& e) {
+        logger_.error("Error while connecting to MQTT broker: {}", e.what());
+      } catch (...) {
```
jbbjarnason, line 110: does this ever occur? because you are catching C any arg exception? 
I think the std::exception is enough

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+    for (auto& slot : slots_) {
+      std::visit([](auto& ptr) { ptr->cancel(); }, slot);
+    }
+
+    active_signals_.clear();
+    get_signals();
+    clean_signals();
+
+    for (tfc::ipc_ruler::signal& signal : active_signals_) {
+      handle_signal(signal);
```
jbbjarnason, line 133: This is okay and could be misunderstood, when reading the code it is nice to know what is going on and `handle_signal` does not describe what will occur. 
I would propose make something like:
```c++
auto slot = make_connection(signal);
slots_.emplace_back(slot);
asio::co_spawn(slot...);
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+                                                                      [&](const std::string& banned_string) {
+                                                                        return signal.name.find(banned_string) !=
+                                                                               std::string::npos;
+                                                                      });
+                                         }),
+                          active_signals_.end());
+  }
+
+  // This function checks the type of the signal in order to determine how to read from the slot
+  auto handle_signal(tfc::ipc_ruler::signal& signal) -> void {
```
jbbjarnason, line 161: we have similar code in impl.hpp of ipc, could we reuse that code or change it somehow to being more reusable

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+    slots_.push_back(std::make_shared<details::slot<slot_type>>(ctx_, signal.name));
+    auto& slot = std::get<std::shared_ptr<details::slot<slot_type>>>(slots_.back());
+    slot->connect(signal.name);
+    asio::co_spawn(mqtt_client_->strand(), slot_coroutine<slot_type, value_type>(*slot, signal.name), asio::detached);
+  }
+
+  // This function stops the coroutines that are currently running and then restarts them
+  auto restart() -> void {
+    stop_coroutine_ = true;
+    ctx_.run_for(std::chrono::milliseconds(20));
```
jbbjarnason, line 205: why is this needed?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+    logger_.info("Signal list has been updated, reloading all signals");
+    load_signals();
+    stop_coroutine_ = false;
+  }
+
+  // This function is called when a signal is received
+  void update_signals([[maybe_unused]] sdbusplus::message::message& msg) noexcept { restart(); }
```
jbbjarnason, line 212: this function is marked `noexcept` but not restart, is restart `noexcept`?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+
+  // This function is called when a value is received and its value needs to be converted to a string
+  template <typename value_type>
+  auto convert_to_string(value_type value) -> std::string {
```
jbbjarnason, line 216: how about using `fmt::format("{}", value)` instead of this function?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+  ipc_client_type ipc_client_;
+  std::shared_ptr<mqtt_client_type> mqtt_client_;
+  tfc::logger::logger logger_;
+  std::unique_ptr<sdbusplus::asio::connection, std::function<void(sdbusplus::asio::connection*)>> dbus_connection_;
+  std::unique_ptr<sdbusplus::bus::match::match, std::function<void(sdbusplus::bus::match::match*)>> signal_updates_;
+  tfc::confman::config<config> config_;
+
+  std::vector<std::string> banned_signals_;
+  std::vector<tfc::ipc_ruler::signal> active_signals_;
+  std::vector<std::variant<std::shared_ptr<details::slot<details::type_bool>>,
```
jbbjarnason, line 295: maybe extract the variant with `using`

exes/mqtt-broadcaster/src/main.cpp
```diff
@@ -0,0 +1,43 @@
+#include <async_mqtt/all.hpp>
+#include <boost/program_options.hpp>
+#include <tfc/ipc.hpp>
+#include "mqtt_broadcaster.hpp"
+
+auto main(int argc, char* argv[]) -> int {
+  try {
+    auto program_description{ tfc::base::default_description() };
+
+    std::string mqtt_host;
```
jbbjarnason, line 10: I think it might be easier to edit the host via config interface, as this will be a common setting to edit

exes/mqtt-broadcaster/src/main.cpp
```diff
@@ -0,0 +1,43 @@
...
+    std::string mqtt_port;
```
jbbjarnason, line 11: make default port the default of mqtt

exes/mqtt-broadcaster/src/main.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+    tfc::base::init(argc, argv, program_description);
+
+    asio::io_context ctx{};
+
+    tfc::ipc_ruler::ipc_manager_client ipc_client{ ctx };
+
+    const std::shared_ptr<async_mqtt::endpoint<async_mqtt::role::client, async_mqtt::protocol::mqtt>> mqtt_client =
+        std::make_shared<async_mqtt::endpoint<async_mqtt::role::client, async_mqtt::protocol::mqtt>>(
+            async_mqtt::protocol_version::v5, ctx.get_executor());
```
jbbjarnason, line 29: Can `ipc_client` and `mqtt_client` be constructed within mqtt_broadcaster?

exes/mqtt-broadcaster/src/main.cpp
```diff
@@ -0,0 +1,43 @@
...
+                           async_mqtt::endpoint<async_mqtt::role::client, async_mqtt::protocol::mqtt>>
+        application(ctx, mqtt_host, mqtt_port, mqtt_username, mqtt_password, ipc_client, mqtt_client);
+
+    ctx.run();
+
+  } catch (std::exception& e) {
+    std::cerr << "Exception caught in main: " << e.what() << "\n";
+  } catch (...) {
+    std::cerr << "Unknown exception caught in main\n";
+  }
```
jbbjarnason, line 41: any specific reason for this exception mechanism?

exes/mqtt-broadcaster/tests/mqtt_broadcaster_tests.cpp
```diff
@@ -0,0 +1,294 @@
+#include <async_mqtt/all.hpp>
+#include <boost/ut.hpp>
+#include <tfc/ipc.hpp>
+#include "mqtt_broadcaster.hpp"
+
+namespace ut = boost::ut;
+using boost::ut::operator""_test;
+
+const int TIMEOUT_IN_MS = 100;
```
jbbjarnason, line 9: I don't like to mix linguistics from C in C++ so I would recommend against using capital casing. 
```suggestion
constexpr int duration_100ms = 100;
```
It would be awesome if std::chrono::duration constructor would be constexpr ...

exes/mqtt-broadcaster/tests/mqtt_broadcaster_tests.cpp
jbbjarnason, line 1: I would suggest making unit tests for each part of the code, but this probably covers the use cases but it is hard to read through the tests.

exes/mqtt-broadcaster/tests/mqtt_broadcaster_tests.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+  tfc::ipc::signal<tfc::ipc::details::type_bool, tfc::ipc_ruler::ipc_manager_client_mock> sig(ctx, ipc_client, "test_signal",
+                                                                                              "description");
+
+  const std::shared_ptr<mock_mqtt_client> mqtt_client =
+      std::make_shared<mock_mqtt_client>(async_mqtt::protocol_version::v5, ctx.get_executor());
+
+  const mqtt_broadcaster<tfc::ipc_ruler::ipc_manager_client_mock&, mock_mqtt_client> application(
+      ctx, std::move(mqtt_host), std::move(mqtt_port), std::move(mqtt_username), std::move(mqtt_password), ipc_client,
+      mqtt_client);
```
jbbjarnason, line 125: this part should be reused instead of duplicated

exes/mqtt-broadcaster/tests/mqtt_broadcaster_tests.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+  ctx.run_for(std::chrono::milliseconds(TIMEOUT_IN_MS));
```
jbbjarnason, line 127: it is pretty long to wait 100 ms often in each test case. how could we make the wait shorter?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+
+namespace asio = boost::asio;
+namespace details = tfc::ipc::details;
+
+// File under /etc/tfc/mqtt-broadcaster/def/mqtt_broadcaster.json which lists the signals that are not supposed to be
+// published on the mqtt broker. Updating the signals in this file will cause the running program to stop and restart
+struct config {
+  tfc::confman::observable<std::vector<std::string>> _banned_topics{};
+  struct glaze {
+    static constexpr auto value{ glz::object("_banned_topics", &config::_banned_topics) };
```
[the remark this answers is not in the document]
magni-mar: Good point, will change this to allowed topics.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: I tried that initially, but the problem is that after a new config has arrived on the observable it isn't loaded immediatly. So when the observable is called:

```cpp
   // When a new list of allowed string arrives the program is restarted
    config_.value()._allowed_topics.observe([this]([[maybe_unused]] auto& new_conf, [[maybe_unused]] auto& old_conf) {
      restart();
    });
```

and consequently clean_signals is called:

```cpp
  // This function filter out signals that are not allowed
  auto clean_signals() -> void {
    active_signals_.erase(std::remove_if(active_signals_.begin(), active_signals_.end(),
                                         [&](const tfc::ipc_ruler::signal& signal) {
                                           return std::ranges::none_of(config_.value()._allowed_topics.value().begin(),
                                                                       config_.value()._allowed_topics.value().end(),
                                                                       [&](const std::string& allowed_string) {
                                                                         return signal.name.find(allowed_string) !=
                                                                                std::string::npos;
                                                                       });
                                         }),
                          active_signals_.end());
  }
```

the value of 

```cpp
config_.value()._allowed_topics.value().begin()
```

is still the old configuration. Then on the next update the previous one is broadcasted.

magni-mar: I tried that initially, but the problem is that after a new config has arrived on the observable it isn't loaded immediatly. So when the observable is called:

```cpp
   // When a new list of allowed string arrives the program is restarted
    config_.value()._allowed_topics.observe([this]([[maybe_unused]] auto& new_conf, [[maybe_unused]] auto& old_conf) {
      restart();
    });
```

and consequently clean_signals is called:

```cpp
  // This function filter out signals that are not allowed
  auto clean_signals() -> void {
    active_signals_.erase(std::remove_if(active_signals_.begin(), active_signals_.end(),
                                         [&](const tfc::ipc_ruler::signal& signal) {
                                           return std::ranges::none_of(config_.value()._allowed_topics.value().begin(),
                                                                       config_.value()._allowed_topics.value().end(),
                                                                       [&](const std::string& allowed_string) {
                                                                         return signal.name.find(allowed_string) !=
                                                                                std::string::npos;
                                                                       });
                                         }),
                          active_signals_.end());
  }
```

the value of 

```cpp
config_.value()._allowed_topics.value().begin()
```

is still the old configuration. Then on the next update the previous one is broadcasted.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: How about a MAC address or a randomly generated hash?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: > it is not really a multi thread problem, this routine can only be spawned once.

How about changing the comment to:

```cpp
  // Connect the mqtt client to the broker. This function can only be spawned once to avoid multiple spawns to open/close the
  // socket at the same time. The routine that calls this function must set the connect_active_ flag to true before calling
  // this function and the function sets it to false when finished.
```

magni-mar: > you probably could change the connect flag to a strand to make this safe

not sure how to do this but will look into it. would appreciate suggestions.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: find it unlikely, changing to only std::exception.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+  }
+
+  auto get_signals() -> void {
+    ipc_client_.signals([&](const std::vector<tfc::ipc_ruler::signal>& signals) {
+      for (const tfc::ipc_ruler::signal& signal : signals) {
+        active_signals_.push_back(signal);
+      }
+    });
+
+    ctx_.run_for(std::chrono::milliseconds(20));
```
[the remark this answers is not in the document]
magni-mar: if this line is skipped then the signals do not load

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: Do you propose this change inside of run_slot, handle_signal or load_signals? the handle_signal function is at least needed to determine the type.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
jbbjarnason: you could use asio post

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: I cannot find similar code in impl.hpp, should this function be moved to be inside of impl.hpp?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+    } else if constexpr (std::is_arithmetic_v<value_type>) {
+      return std::to_string(value);
+    } else {
+      return "unknown";
+    }
+  }
+
+  // This function runs a coroutine for a slot, when a new value is received a message is sent to the MQTT broker
+  template <class slot_type, class value_type>
+  auto slot_coroutine(details::slot<slot_type>& slot, std::string signal_name) -> asio::awaitable<void> {
```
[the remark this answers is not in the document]
magni-mar: the parameter in the function?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: If this variable is left out and the ctx isn't allowed to run then the old slots are not cancelled.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: noexcept can be skipped in this function

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: that's a better solution, deleting this function.

exes/mqtt-broadcaster/src/main.cpp
magni-mar: The exception mechanism was suggested by clang-tidy because the main function could throw an exception, taking this out.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
jbbjarnason: > > you probably could change the connect flag to a strand to make this safe
> 
> not sure how to do this but will look into it. would appreciate suggestions.

if the class would own a connecting_strand member you could post the connect method on the strand and in the beginning of the function check if you are already connected

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
jbbjarnason: how about the hostname?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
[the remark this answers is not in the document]
jbbjarnason: then there is something wrong here because you run the context in the main.cpp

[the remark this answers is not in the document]
jbbjarnason: it seems like you are using the `signals` function as synchronous call when it is asynchronous so you are blocking here for 20 ms so the user of this function will not continue until that has exceeded

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
jbbjarnason: a similar behavior is on the bottom of the impl.hpp

exes/mqtt-broadcaster/src/main.cpp
magni-mar: I can move mqtt_client but the ipc_client is needed for testing.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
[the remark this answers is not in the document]
jbbjarnason: yes owning stack reference to a variable which has been constructed as pointer can lead to undefined behavior. however if you are confident that the lifetime of the object always exceeds the lifetime of this coroutine  you are good.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
jbbjarnason: I mean the run_for, is it the same as the other one? blocking the consumer of the restart function?

exes/mqtt-broadcaster/src/main.cpp
jbbjarnason: you can use different constructor for testing

exes/mqtt-broadcaster/tests/mqtt_broadcaster_tests.cpp
magni-mar: Changed.

exes/mqtt-broadcaster/tests/mqtt_broadcaster_tests.cpp
magni-mar: lowering it to 25 ms

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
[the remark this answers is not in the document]
magni-mar: yes that is what I am doing, right now I am trying to find a way around this because this is unreliable behaviour.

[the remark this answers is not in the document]
magni-mar: will probably include the cleaning of signals and the rest of the load_signals function inside the callback

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: after refactoring other parts of the code neither asio::post or this variable is necessary, simply accessing the value is enough.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: local ip number or public ip? local would probaby be more beneficial since these hosts are most often running on the same LAN.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,305 @@
...
+      // If the connection failed, then clear the socket and try again after 1 second.
+      logger_.trace("Waiting for 1 second and then trying to connect again");
+      co_await asio::co_spawn(mqtt_client_->strand(), mqtt_client_->close(asio::use_awaitable), asio::use_awaitable);
+      co_await asio::steady_timer{ ctx_, std::chrono::seconds{ 1 } }.async_wait(asio::use_awaitable);
+    }
+  }
+
+  // Function which has oversight over the signals
+  auto load_signals() -> void {
+    // Stop reading the current signals by canceling the slots
```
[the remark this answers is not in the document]
magni-mar: only adding signals that have been added, signals can not be removed

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: this code has changed significantly so marking this as resolved.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: this has changed, only new signals are added.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: this code has changed significantly and types are declared automatically, so this issue has been resolved.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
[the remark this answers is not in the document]
magni-mar: although this code has changed the principle of your comment has been implemented in the new version. A vector of pointer is owned by the class to ensure that its lifetime is longer than the function call.

exes/mqtt-broadcaster/src/main.cpp
magni-mar: you are right, moving the construction of mqtt_client and ipc_client to the normal constructor instead of being defined before.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: > > > it is not really a multi thread problem, this routine can only be spawned once.
> > 
> > 
> > How about changing the comment to:
> > ```c++
> >   // Connect the mqtt client to the broker. This function can only be spawned once to avoid multiple spawns to open/close the
> >   // socket at the same time. The routine that calls this function must set the connect_active_ flag to true before calling
> >   // this function and the function sets it to false when finished.
> > ```
> 
> seems okay, but why is the flag not set in the function itself, moving the responsibility of the flag to the function

done

exes/mqtt-broadcaster/src/main.cpp
magni-mar: moved this to config interface along with allowed_topics

magni-mar:
Closing this PR because more features need to be added to the executable until it can be merged, such as SparkPlug B and more.

PR #118, opened 2023-07-06
Adding properties callback on dbus

magni-mar:
This change allows an outside class to define a matcher which will invoke a callback whenever a property is changed on the D-bus. 

The callback is then owned by the caller.

The mock client was changed in accordance.

### DIFF
--- a/libs/ipc/inc/public/tfc/ipc.hpp
+++ b/libs/ipc/inc/public/tfc/ipc.hpp
@@ -74,8 +74,6 @@ public:
 
   [[nodiscard]] auto value() const noexcept { return slot_->get(); }  // todo: value() or get()
 
-  auto something() -> std::string { return "something"; }
-
 private:
   static constexpr std::string_view self_name{ slot_tag };
   std::shared_ptr<details::slot_callback<type_desc>> slot_;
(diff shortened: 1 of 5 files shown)


libs/ipc/inc/public/tfc/ipc.hpp
```diff
@@ -73,6 +74,8 @@ class slot {
 
   [[nodiscard]] auto value() const noexcept { return slot_->get(); }  // todo: value() or get()
 
+  auto something() -> std::string { return "something"; }
+
```
omarhogni, line 78: remove.

libs/ipc/inc/public/tfc/ipc/details/dbus_server_iface.hpp
```diff
@@ -283,13 +294,19 @@ class ipc_manager_server {
   std::unique_ptr<sdbusplus::asio::dbus_interface> dbus_interface_;
   std::unique_ptr<sdbusplus::asio::object_server> object_server_;
   std::unique_ptr<ipc_manager<signal_storage, slot_storage>> ipc_manager_;
+
+  const char* connection_change = "ConnectionChange";
```
omarhogni, line 298: Please define in the same manor as the other dbus interfaces.

PR #128, opened 2023-07-25
MQTT broadcaster which follows the Spark Plug B specification.

magni-mar:
Reads values of signals and forwards those values to an MQTT broker that follows the Spark Plug B specification.

Uses MQTT v5.

https://sparkplug.eclipse.org/specification/version/3.0/documents/sparkplug-specification-3.0.0.pdf


exes/mqtt-broadcaster/inc/config.hpp
```diff
@@ -0,0 +1,42 @@
...
+                                             &config::_mqtt_broker_password,
+
+                                             "_sparkplug_b_edge_node_id",
+                                             &config::_sparkplug_b_edge_node_id,
+
+                                             "_sparkplug_b_group_id",
+                                             &config::_sparkplug_b_group_id,
+
+                                             "_client_id",
+                                             &config::_client_id
```
jbbjarnason, line 37: Why the underscore prefix?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,560 @@
...
+namespace asio = boost::asio;
+namespace details = tfc::ipc::details;
+
+class network_manager {
+public:
+  auto connect_socket(auto&& socket, auto&& resolved_ip) -> asio::awaitable<void> {
+    co_await asio::async_connect(socket, resolved_ip, asio::use_awaitable);
+  }
+
+  auto set_sni_hostname(auto&& native_handle, std::string broker_address) -> void {
```
jbbjarnason, line 23: passing std::string by value is not preferred even though the string is small, change to `std::string const& broker_address`

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,560 @@
...
+        logger_("mqtt_broadcaster"), config_(io_ctx, "mqtt_broadcaster"), network_manager_() {
+    properties_callback_ = ipc_client_.register_properties_change_callback(
+        [this]([[maybe_unused]] sdbusplus::message_t& msg) { add_new_signals(); });
+    asio::co_spawn(mqtt_client_->strand(), initialize(), asio::detached);
+  }
+
+  explicit mqtt_broadcaster(asio::io_context& io_ctx, bool)
+      : io_ctx_(io_ctx), tls_ctx_{ async_mqtt::tls::context::tlsv12 }, ipc_client_{ io_ctx_ },
+        mqtt_client_(std::make_shared<mqtt_client_type>(async_mqtt::protocol_version::v5, io_ctx.get_executor(), tls_ctx_)),
+        logger_("mqtt_broadcaster"), config_(io_ctx, "mqtt_broadcaster"), network_manager_() {}
```
jbbjarnason, line 64: what is this constructor doing? It accepts an unused boolean

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,560 @@
...
+  auto get_payload() -> Payload {
+    Payload payload;
+    payload.set_timestamp(timestamp_milliseconds().count());
+    if (seq_ == 255) {
+      seq_ = 0;
+    } else {
+      seq_++;
+    }
+
+    payload.set_seq(seq_);
```
jbbjarnason, line 89: maybe rename function to make_payload as the operation `get` is antonym of `set` here

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,560 @@
...
+      case details::type_e::_json:
+        return 12;
+    }
+    return 0;
+  }
+
+  auto format_signal_name(std::string signal_name_to_format) -> std::string {
+    while (signal_name_to_format.find('.') != std::string::npos) {
+      signal_name_to_format.replace(signal_name_to_format.find('.'), 1, "/");
+    }
```
jbbjarnason, line 116:
```suggestion
    std::replace(signal_name_to_format.begin(), signal_name_to_format.end(), '.', '/');
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,560 @@
...
+
+template <class ipc_client_type, class mqtt_client_type, class config_type, class network_manager_type>
+class mqtt_broadcaster {
+public:
+  static constexpr std::string_view namespace_element = "spBv1.0";
+  static constexpr std::string_view ndata = "NDATA";
+  static constexpr std::string_view nbirth = "NBIRTH";
+  static constexpr std::string_view ndeath = "NDEATH";
+  static constexpr std::string_view ncmd = "NCMD";
+  static constexpr std::string_view rebirth_metric = "Node Control/Rebirth";
```
jbbjarnason, line 50: are these variables not available from proto definition?

exes/mqtt-broadcaster/inc/config.hpp
magni-mar: My thinking was that it might be a good idea to have a specific naming convention for variables that are kept in a configuration file so that they are easily identifiable.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: The unused boolean is a subpar solution but the point of this constructor is that unit tests can use it and skip the following lines:

```cpp
properties_callback_ = ipc_client_.register_properties_change_callback([this]([[maybe_unused]]sdbusplus::message_t& msg) { add_new_signals(); });
asio::co_spawn(mqtt_client_->strand(), initialize(), asio::detached);
```

which prevents the class from starting to run the initialization function, allowing the tests to run each function individually.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,560 @@
...
+    } else {
+      seq_++;
+    }
+
+    payload.set_seq(seq_);
+
+    return payload;
+  }
+
+  auto type_enum_convert(tfc::ipc::details::type_e type) -> uint32_t {
```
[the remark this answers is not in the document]
magni-mar: The come from the specification

https://sparkplug.eclipse.org/specification/version/3.0/documents/sparkplug-specification-3.0.0.pdf

adding a comment above the function to make this clear.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: They are not to be found in the proto header file at least. Have not found a way to retrieve them.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,560 @@
...
+
+  auto timestamp_milliseconds() -> std::chrono::milliseconds {
+    using std::chrono::duration_cast;
+    using std::chrono::milliseconds;
+    using std::chrono::system_clock;
+    return duration_cast<milliseconds>(system_clock::now().time_since_epoch());
+  }
+
+  auto get_payload() -> Payload {
+    Payload payload;
```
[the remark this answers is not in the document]
magni-mar: Doesn't seem to be available.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,560 @@
...
+      logger_.trace("Waiting for NCMD packet...");
+
+      auto publish_packet_received = co_await mqtt_client_->recv(
+          async_mqtt::filter::match, { async_mqtt::control_packet_type::publish }, asio::use_awaitable);
+
+      auto publish_packet = publish_packet_received.template get<async_mqtt::v5::publish_packet>();
+
+      logger_.trace("Received PUBLISH packet. Parsing payload...");
+
+      auto data = publish_packet.payload()[0];
```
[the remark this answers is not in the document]
magni-mar: I can add this line to make it safe:

```cpp
if (publish_packet.payload().size() == 0) {
    logger_.warn("Received invalid payload. Cannot process packet.");
    co_return;
}
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,560 @@
...
+
+      Payload payload;
+
+      bool payload_valid = payload.ParseFromArray(data.data(), data.size());
+
+      if (!payload_valid) {
+        logger_.warn("Received invalid payload. Continuing to next iteration...");
```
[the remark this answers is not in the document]
magni-mar: As taken from the relevant documentation:

```cpp
  // Fill the message with a protocol buffer parsed from the given input
  // stream. Returns false on a read error or if the input is in the wrong
  // format.  A successful return does not indicate the entire input is
  // consumed, ensure you call ConsumedEntireMessage() to check that if
  // applicable.
```

It will only return false on error, not the relevant error.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
jbbjarnason: I would recommend either to make a comment on this constructor

Or, remove this constructor, make initialize do everything that is needed or extract new function `run` and call that function within main.

magni-mar: > make initialize do everything that is needed

That would just be the normal constructor, which has a lot of side effects from the initialization function.

> extract new function `run` and call that function within main.

If the constructor calls `run` in both testing and normal execution then I still have the same problem.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,560 @@
...
+namespace details = tfc::ipc::details;
+
+class network_manager {
+public:
+  auto connect_socket(auto&& socket, auto&& resolved_ip) -> asio::awaitable<void> {
+    co_await asio::async_connect(socket, resolved_ip, asio::use_awaitable);
+  }
+
+  auto set_sni_hostname(auto&& native_handle, std::string broker_address) -> void {
+    if (!SSL_set_tlsext_host_name(native_handle, broker_address.c_str())) {
```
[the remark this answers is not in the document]
magni-mar: The function is setting the SNI of the socket, I have not found how to do this using openssl.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,618 @@
...
+
+  auto set_value_payload(Payload_Metric* metric, int64_t& value) -> void { metric->set_long_value(value); }
+
+  auto set_value_payload(Payload_Metric* metric, double& value) -> void { metric->set_double_value(value); }
+
+  auto set_value_payload(Payload_Metric* metric, float& value) -> void { metric->set_float_value(value); }
+
+  auto set_value_payload(Payload_Metric* metric, uint32_t& value) -> void { metric->set_int_value(value); }
+
+  auto connect_to_broker() -> asio::awaitable<void> {
```
omarhogni, line 169: I would like to see you clean up this method.
try not using throw catch as goto's.

You can use asio::as_tuple to get rid of the exceptions and
get a cleaner control flow inside the method. You can also compose some async operations
like "connect_and_handshake" and hide the implementation details away from
the reconnecting logic.

I would also say some of these operations dont belong inside the loop. Why do we have to set_default_verify_paths for the ssl context each time a connect fails? or the verify mode.

This could also be seen to be as broad because you dont consider which operation failed.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,618 @@
...
+  explicit mqtt_broadcaster(asio::io_context& io_ctx, bool)
+      : io_ctx_(io_ctx), tls_ctx_{ async_mqtt::tls::context::tlsv12 }, ipc_client_{},
+        mqtt_client_(std::make_shared<mqtt_client_type>(async_mqtt::protocol_version::v5, io_ctx.get_executor(), tls_ctx_)),
+        logger_("mqtt_broadcaster"), config_(io_ctx, "mqtt_broadcaster"), network_manager_(), timer_{ io_ctx_ } {}
+
+private:
+  auto initialize() -> asio::awaitable<void> {
+    co_await asio::co_spawn(mqtt_client_->strand(), connect_to_broker(), asio::use_awaitable);
+    add_new_signals();
+    asio::co_spawn(mqtt_client_->strand(), ncmd_listener(), asio::detached);
```
omarhogni, line 71: Why co_spawn here, you just co_spawned initalize and marked it detached.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,618 @@
...
+  auto ncmd_listener() -> asio::awaitable<void> {
+    co_await subscribe_to_ncmd_topic();
+
+    while (true) {
+      try {
+        co_await process_ncmd_packet();
+      } catch (std::exception const&) {
+        break;
+      }
+    }
```
omarhogni, line 250: All this method does is destroy the error information produced inside process_ncmd_packet
why do we not care about this error?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,618 @@
...
+      auto publish_packet = publish_packet_received.template get<async_mqtt::v5::publish_packet>();
+
+      logger_.trace("Received PUBLISH packet. Parsing payload...");
+
+      if (publish_packet.payload().size() == 0) {
+        logger_.warn("Received invalid payload. Cannot process packet.");
+        co_return;
+      }
+
+      auto data = publish_packet.payload()[0];
```
omarhogni, line 269: Can there not be two payloads?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,618 @@
...
+      *p_id, { { async_mqtt::allocate_buffer(ncmd_topic), async_mqtt::qos::at_most_once | async_mqtt::sub::nl::yes } }
+    };
+
+    logger_.trace("Sending subscription packet...");
+
+    co_await mqtt_client_->send(subscribe_packet, asio::use_awaitable);
+
+    logger_.trace("Subscription packet sent. Waiting for SUBACK...");
+
+    co_await mqtt_client_->recv(async_mqtt::filter::match, { async_mqtt::control_packet_type::suback }, asio::use_awaitable);
```
omarhogni, line 324: Does this not have a return code that needs to be checked?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,618 @@
...
+                if (error_code) {
+                  logger_.trace("Error connecting to signal: {}, error: {}", signal.name, error_code.message());
+                }
+              }
+            },
+            ipc);
+
+        signal_data signal_data{ signal, std::move(ipc), std::nullopt };
+
+        signals_.emplace_back(std::move(signal_data));
```
omarhogni, line 378: you can emplace straight away and save the boiler plate.
```suggestion
        signals_.emplace_back(signal, std::move(ipc), std::nullopt);
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,618 @@
...
+          },
+          signal_data.receiver);
+
+      variable_metric->set_is_transient(false);
+      variable_metric->set_is_historical(false);
+
+      logger_.info("Signal_data processed and added to the payload: {}", signal_data.information.name);
+    }
+
+    payload.PrintDebugString();
```
omarhogni, line 504: All console IO should use the logger.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,618 @@
...
+      send_failed = true;
+    }
+
+    if (send_failed) {
+      logger_.info("Attempting to reconnect to broker");
+      co_await asio::co_spawn(mqtt_client_->strand(), connect_to_broker(), asio::use_awaitable);
+      logger_.info("Reconnected to broker. Attempting to resend the message to topic: {}", topic);
+      co_await send_message(topic, payload, _qos);
+    }
+  }
```
omarhogni, line 591: Can you not just use the catch?
```suggestion

    try {
      uint16_t packet_id = 0;
      if (_qos != async_mqtt::qos::at_most_once) {
        packet_id = mqtt_client_->acquire_unique_packet_id().value();
      }
      logger_.info("Attempting to send message to topic: {}", topic);
      auto result =
          co_await mqtt_client_->send(async_mqtt::v5::publish_packet{ packet_id, async_mqtt::allocate_buffer(topic),
                                                                      async_mqtt::allocate_buffer(payload), _qos },
                                      asio::use_awaitable);

      if (result) {
        logger_.error("Failed to send message to topic: {}", topic);
        send_failed = true;
      } else {
        logger_.info("Message successfully sent to topic: {}", topic);
      }
    } catch (std::exception const& exc) {
      logger_.error("Exception occurred while sending message to topic: {}. Exception: {}", topic, exc.what());
      logger_.info("Attempting to reconnect to broker");
      co_await asio::co_spawn(mqtt_client_->strand(), connect_to_broker(), asio::use_awaitable);
      logger_.info("Reconnected to broker. Attempting to resend the message to topic: {}", topic);
      co_await send_message(topic, payload, _qos);
    }
  }
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,618 @@
...
+      const boost::system::error_code error{ static_cast<int>(::ERR_get_error()), boost::asio::error::get_ssl_category() };
+      throw boost::system::system_error{ error };
+    }
+  }
+
+  auto async_handshake(auto&& socket) -> asio::awaitable<void> {
+    co_await std::forward<decltype(socket)>(socket).async_handshake(async_mqtt::tls::stream_base::client,
+                                                                    asio::use_awaitable);
+  }
+};
```
omarhogni, line 35: I am not really a fan of this but don't have an easy solution that
works instead.

omarhogni (changes requested):

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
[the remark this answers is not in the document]
omarhogni: @magni-mar he means that you are not including the header the method is from.
You get it from async-mqtt but since you are using the method directly you should include the openssl header. Specifically ``` #include <openssl/ssl.h>``` I think.

exes/mqtt-broadcaster/inc/config.hpp
magni-mar: changing this to be cleaner.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: Constructor has been removed and a run() function has been added which does the same job as the constructor along with running the io ctx. The tests would ignore this function when unit testing. That way the callee only needs to call:

```cpp
mqtt_broadcaster<tfc::ipc_ruler::ipc_manager_client, async_mqtt::endpoint<async_mqtt::role::client, async_mqtt::protocol::mqtts>, tfc::confman::config<config>, network_manager> application(io_ctx);

application.run();
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,618 @@
...
+        logger_("mqtt_broadcaster"), config_(io_ctx, "mqtt_broadcaster"), network_manager_(), timer_{ io_ctx_ } {
+    properties_callback_ = ipc_client_.register_properties_change_callback(
+        [this]([[maybe_unused]] sdbusplus::message_t& msg) { add_new_signals(); });
+    asio::co_spawn(mqtt_client_->strand(), initialize(), asio::detached);
+  }
+
+  explicit mqtt_broadcaster(asio::io_context& io_ctx, bool)
+      : io_ctx_(io_ctx), tls_ctx_{ async_mqtt::tls::context::tlsv12 }, ipc_client_{},
+        mqtt_client_(std::make_shared<mqtt_client_type>(async_mqtt::protocol_version::v5, io_ctx.get_executor(), tls_ctx_)),
+        logger_("mqtt_broadcaster"), config_(io_ctx, "mqtt_broadcaster"), network_manager_(), timer_{ io_ctx_ } {}
```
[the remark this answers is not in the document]
magni-mar: Already resolved.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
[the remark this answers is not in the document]
magni-mar: https://www.openssl.org/docs/man1.1.1/man3/SSL_set_tlsext_host_name.html

Ahh okey, I misunderstood, will include the header.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: I refactored the function into two subfunctions and changed the control flow. Please review the new function when possible.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: You are right, it is unnecessary. Changing this to 

```cpp
auto initialize() -> asio::awaitable<void> {
    co_await connect_to_broker();
    add_new_signals();
    co_await ncmd_listener();
}
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: I will add a log statement so that the error is used.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: I don't see why not. Changing this to a for loop and processing each payload in another function, also adds readability.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: The return value from the function is a suback packet and we can check its reason code:

https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901174

The only question is what should be done with the information in the packet, should the program be killed? If the function is returned then the caller will think that the subscribe was successful and will wait for a packet.

```cpp
auto suback_received = co_await mqtt_client_->recv(async_mqtt::filter::match, { async_mqtt::control_packet_type::suback }, asio::use_awaitable);

auto suback_packet = suback_received.template get<async_mqtt::v5::suback_packet>();

for (auto const& entry : suback_packet.entries()) {
    if (entry != async_mqtt::suback_reason_code::granted_qos_0) {
    logger_.error("Error subscribing to topic: {}, reason code: {}", ncmd_topic, async_mqtt::suback_reason_code_to_str(entry));
    std::exit(-1);
    }
}
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: Changing to:

```cpp
logger_.trace("{}", payload.DebugString());
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: If the result from the send is not successful then a reconnect won't happen.

```cpp
if (result) {
    logger_.error("Failed to send message to topic: {}", topic);
    send_failed = true;
}
```

this doesn't result in a reconnect.

magni-mar: after reading the send function in async_mqtt I don't think it throws an exception. This should be enough:

```cpp
auto send_message(std::string topic, std::string payload, async_mqtt::qos _qos) -> asio::awaitable<void> {
    uint16_t packet_id = 0;
    if (_qos != async_mqtt::qos::at_most_once) {
        packet_id = mqtt_client_->acquire_unique_packet_id().value();
    }
    
    logger_.info("Attempting to send message to topic: {}", topic);

    auto error_code = co_await mqtt_client_->send(async_mqtt::v5::publish_packet{ packet_id, async_mqtt::allocate_buffer(topic), async_mqtt::allocate_buffer(payload), _qos }, use_nothrow_awaitable);

    if (error_code) {
        logger_.error("Failed to send message to topic: {}", topic);
        logger_.info("Attempting to reconnect to broker");
      
        co_await asio::co_spawn(mqtt_client_->strand(), connect_to_broker(), asio::use_awaitable);
        logger_.info("Reconnected to broker. Attempting to resend the message to topic: {}", topic);
        co_await send_message(topic, payload, _qos);
    } else {
        logger_.info("Message successfully sent to topic: {}", topic);
    }
}
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: agree

exes/mqtt-broadcaster/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+         Boost::log
+         Boost::program_options
+         OpenSSL::SSL OpenSSL::Crypto
+         tfc::sparkplug::proto
+         )
+
+ target_compile_definitions(mqtt-broadcaster-unit-tests
+         PUBLIC
+         ASYNC_MQTT_USE_TLS
+         )
```
jbbjarnason, line 57: formatting of this section does not follow previous work

exes/mqtt-broadcaster/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
+ add_executable(mqtt-broadcaster "src/main.cpp")
+ add_executable(mqtt-broadcaster-unit-tests "tests/unit_tests.cpp")
```
jbbjarnason, line 2: most common pattern in repo is to have CMakeLists file under tests directory, it will be easier to remove targets with BUILD_TESTING cmake flag

jbbjarnason: example 
```
if (BUILD_TESTING)
  add_subdirectory(tests)
endif ()
```

exes/mqtt-broadcaster/inc/config.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <tfc/confman.hpp>
+#include <tfc/confman/observable.hpp>
+
+// File under /etc/tfc/mqtt-broadcaster/def/mqtt_broadcaster.json which lists the signals that are supposed to be published
+// on the mqtt broker. Updating the signals in this file will cause the running program to stop and restart
+struct config {
+  std::string address{};
+  std::string port{};
```
jbbjarnason, line 10: why do we have port as string here instead of uint16_t?

exes/mqtt-broadcaster/inc/config.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <tfc/confman.hpp>
+#include <tfc/confman/observable.hpp>
+
+// File under /etc/tfc/mqtt-broadcaster/def/mqtt_broadcaster.json which lists the signals that are supposed to be published
+// on the mqtt broker. Updating the signals in this file will cause the running program to stop and restart
+struct config {
+  std::string address{};
```
jbbjarnason, line 9: could we transpose asio::endpoint with the config?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,635 @@
+#pragma once
+
+#include <openssl/ssl.h>
```
jbbjarnason, line 3: missing all standard includes for things used in the file. even though the includes are implicitly included now does not mean it will be in the future so please add explicit includes for all used standard library features, such as string_view, utility, optional, any ...

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,635 @@
...
+#include <sparkplug_b/sparkplug_b.pb.h>
+#include <async_mqtt/all.hpp>
+#include <boost/asio.hpp>
+#include <boost/asio/experimental/as_tuple.hpp>
+#include <tfc/dbus/string_maker.hpp>
+
+#include "config.hpp"
+
+using org::eclipse::tahu::protobuf::Payload;
+using org::eclipse::tahu::protobuf::Payload_Metric;
```
jbbjarnason, line 13: I would add `namespace tfc` above these lines, otherwise Payload and Payload_Metric and everything else can collide with any of the other includes above or by the user of this code

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: Changing the function to return an error code so that the callee the reconnects to the broker and the function is called again.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,635 @@
...
+    if (!SSL_set_tlsext_host_name(std::forward<decltype(native_handle)>(native_handle), broker_address.c_str())) {
+      const boost::system::error_code error{ static_cast<int>(::ERR_get_error()), boost::asio::error::get_ssl_category() };
+      throw boost::system::system_error{ error };
+    }
+  }
+
+  auto async_handshake(auto&& socket) -> asio::awaitable<void> {
+    co_await std::forward<decltype(socket)>(socket).async_handshake(async_mqtt::tls::stream_base::client,
+                                                                    asio::use_awaitable);
+  }
```
jbbjarnason, line 37: my guess is that these function can be made static since template injection is used

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,635 @@
...
+
+private:
+  auto initialize() -> asio::awaitable<void> {
+    co_await connect_to_broker();
+    add_new_signals();
+    co_await ncmd_listener();
+  }
+
+  auto timestamp_milliseconds() -> std::chrono::milliseconds {
```
jbbjarnason, line 75: this can be made static

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,635 @@
...
+
+  auto set_value_payload(Payload_Metric* metric, uint64_t& value) -> void { metric->set_long_value(value); }
+
+  auto set_value_payload(Payload_Metric* metric, int64_t& value) -> void { metric->set_long_value(value); }
+
+  auto set_value_payload(Payload_Metric* metric, double& value) -> void { metric->set_double_value(value); }
+
+  auto set_value_payload(Payload_Metric* metric, float& value) -> void { metric->set_float_value(value); }
+
+  auto set_value_payload(Payload_Metric* metric, uint32_t& value) -> void { metric->set_int_value(value); }
```
jbbjarnason, line 167: any reason to have value by reference instead of const ref?

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -0,0 +1,635 @@
...
+
+    logger_.trace("SUBACK received. Checking for errors...");
+
+    auto suback_packet = suback_received.template get<async_mqtt::v5::suback_packet>();
+
+    for (auto const& entry : suback_packet.entries()) {
+      if (entry != async_mqtt::suback_reason_code::granted_qos_0) {
+        logger_.error("Error subscribing to topic: {}, reason code: {}", ncmd_topic,
+                      async_mqtt::suback_reason_code_to_str(entry));
+        std::exit(-1);
```
jbbjarnason: why not throw instead of exit?

exes/mqtt-broadcaster/CMakeLists.txt
magni-mar: Added a Cmake Lists file under the tests folder and the BUILD_TESTING option as well.

exes/mqtt-broadcaster/inc/config.hpp
magni-mar: Because the code also supports port name resolution, f.e. "mqtt" resolved to "1883". Another possible name would be "port_service" or something similar. Change the name?

exes/mqtt-broadcaster/inc/config.hpp
magni-mar: Could you please explain further? I'm not sure I understand.

exes/mqtt-broadcaster/inc/config.hpp
```diff
@@ -0,0 +1,42 @@
...
+                                             "node_id",
+                                             &config::node_id,
+
+                                             "group_id",
+                                             &config::group_id,
+
+                                             "client_id",
+                                             &config::client_id
+
+                                             ) };
```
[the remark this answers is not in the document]
magni-mar: Is the description necessary? I feel like the options are self-explanatory.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: Added includes.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: Like this?

```cpp
namespace asio = boost::asio;
namespace details = tfc::ipc::details;

namespace tfc {

using org::eclipse::tahu::protobuf::Payload;
using org::eclipse::tahu::protobuf::Payload_Metric;
...
```

exes/mqtt-broadcaster/CMakeLists.txt
magni-mar: formatted

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: They can, converting to static.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: Not particularly, I can add const <type>& to the parameter.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: This code has changed.

exes/mqtt-broadcaster/inc/config.hpp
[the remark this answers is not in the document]
magni-mar: Valid point, will change.

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
jbbjarnason: yes

exes/mqtt-broadcaster/inc/config.hpp
magni-mar: `mqtts` is also possible for port 8883.

PR #136, opened 2023-08-24
MQTT broadcaster can send values from SCADA to TFC

magni-mar:


exes/mqtt-broadcaster/inc/config.hpp
```diff
@@ -12,6 +12,7 @@ struct config {
   std::string node_id{};
   std::string group_id{};
   std::string client_id{};
+  std::array<std::tuple<std::string, std::string>, 10> scada_signals{};
```
jbbjarnason, line 15: Couple of things I would like to note here,

1. std::array instead of std::vector?
2. std::tuple instead of struct? You could remove the following description `each signal is a tuple of (signal name, signal type). Allowed types are: _bool, _double_t, _int64_t, _json, _string, _uint64_t` by using a struct.

Example:
```cpp
struct signal_definition {
std::string name{};
tfc::ipc::details::type_e type{};
struct glaze {
// ...
};
};

// in mqtt config struct would be
struct config {
// ...
std::vector<signal_definition> scada_signals{};
};
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -60,6 +60,18 @@ struct signal_data {
   std::optional<std::any> current_value;
 };
 
+struct scada_signal {
+  std::variant<tfc::ipc::bool_signal,
+               tfc::ipc::double_signal,
+               tfc::ipc::int_signal,
+               tfc::ipc::json_signal,
+               tfc::ipc::string_signal,
+               tfc::ipc::uint_signal>
```
jbbjarnason, line 69: I would move this to ipc.hpp
```cpp
using bool_signal = signal<details::type_bool, ipc_ruler::ipc_manager_client>;
using int_signal = signal<details::type_int, ipc_ruler::ipc_manager_client>;
using uint_signal = signal<details::type_uint, ipc_ruler::ipc_manager_client>;
using double_signal = signal<details::type_double, ipc_ruler::ipc_manager_client>;
using string_signal = signal<details::type_string, ipc_ruler::ipc_manager_client>;
using json_signal = signal<details::type_json, ipc_ruler::ipc_manager_client>;
using any_signal = std::variant<bool_signal, int_signal, uint_signal, double_signal, string_signal, json_signal>;
```

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -84,11 +96,51 @@ class mqtt_broadcaster {
...
+    create_scada_signals();
     co_await connect_to_broker();
     add_new_signals();
     co_await ncmd_listener();
   }
 
+  auto create_scada_signals() -> void {
+    auto config_scada_signals = config_.value().scada_signals;
+
+    std::vector<std::string> allowed_types = { "_bool", "_double_t", "_int64_t", "_json", "_string", "_uint64_t" };
```
jbbjarnason, line 108: this will become unnecessary if you use `type_e` as the config parameter, as mentioned in config file

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
```diff
@@ -654,6 +763,8 @@ class mqtt_broadcaster {
 
   asio::cancellation_signal cancel_signal_;
 
+  std::vector<scada_signal> scada_signals;
```
jbbjarnason, line 766: this could be `std::vector<ipc::any_signal>` given the config struct

jbbjarnason (changes requested):
Can be simplified greatly by passing the responsibility of parsing config to the glaze parser

exes/mqtt-broadcaster/inc/config.hpp
magni-mar: I agree std::vector should be used instead of std::array.

Initially I used a struct, similar to what you outlined above but backed off for the following reason.

Implementing the solution given above results in a compile error that originates in "json_schema.hpp". The compile error is:

```bash
error: no matching function for call to ‘get<2>(const glz::tuplet::tuple<std::basic_string_view<char, std::char_traits<char> >, tfc::ipc::details::type_e, std::basic_string_view<char, std::char_traits<char> > >&)’
  251 |         enumeration.attributes.description = std::get<2>(item);
```

This compile error happens when the configuration struct is initialized.

jbbjarnason: This should be possible, so the compiler error just needs to be resolved

exes/mqtt-broadcaster/inc/mqtt_broadcaster.hpp
magni-mar: How would I access the name and the signal in the same iteration when searching for a signal?

exes/mqtt-broadcaster/inc/config.hpp
magni-mar: After some debugging, the following is possible:

```cpp
struct signal_defintion {
  std::string name{};
  std::string type{};
  struct glaze {
    // clang-format off
        static constexpr auto value{ glz::object(
          "name", &signal_defintion::name,
          "type", &signal_defintion::type
        )};
    // clang-format on
  };
};
```

but the defining an internal tfc type for a variable is not possible:

```cpp
struct signal_defintion {
  std::string name{};
  tfc::ipc::details::type_e type{};
  struct glaze {
    // clang-format off
        static constexpr auto value{ glz::object(
          "name", &signal_defintion::name,
          "type", &signal_defintion::type
        )};
    // clang-format on
  };
};
```

magni-mar: This now uses tfc types

PR #147, opened 2023-09-06
Cloudsmith

omarhogni:


omarhogni:
Placement looks correct, branch rebased
Screenshot from 2023-10-03 19-09-45

.github/workflows/compile.yml
```diff
@@ -8,6 +8,11 @@ on:
 
 env:
   CONFIGURATION_DIRECTORY: ${{ github.workspace }}
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  VCPKG_FEATURE_FLAGS: dependencygraph
```
jbbjarnason, line 12: what is `dependencygraph` used for?

.github/workflows/compile.yml
```diff
@@ -8,6 +8,11 @@ on:
...
+
+permissions:
+  contents: write
```
jbbjarnason, line 15: where are we writing now?

.github/workflows/ubuntu-packaging.yml
```diff
@@ -72,39 +60,33 @@ jobs:
...
-      - name: Upload rpm package
-        if: matrix.architecture == 'aarch64'
-        uses: actions/upload-artifact@v3
-        with:
-          name: tfc-framework-${{matrix.build_type}}
-          path: build/pack-aarch64-gcc-${{matrix.build_type}}/*.rpm
-      - name: Upload tar package
-        if: matrix.architecture == 'aarch64'
-        uses: actions/upload-artifact@v3
+      - name: Cloudsmith upload
```
jbbjarnason, line 80: should we restrict to only run on master?

cmake/tfc_cpack.cmake
```diff
@@ -26,22 +22,12 @@ set(CPACK_PACKAGE_VERSION_PATCH ${PROJECT_VERSION_PATCH})
...
 #set(CMAKE_INSTALL_PREFIX "/usr" CACHE PATH "Installation prefix" FORCE)
-set(CPACK_PACKAGING_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX})
+#set(CPACK_PACKAGING_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX})
 
 if(CMAKE_STRIP)
   set(CPACK_RPM_SPEC_MORE_DEFINE "%define __strip ${CMAKE_STRIP}")
 endif()
 
-# set dependencies to hosts libc version and cockpit for UI
-if (LDD_RESULT EQUAL 0)
```
jbbjarnason, line 36: I would like to fix this instead of removing, it is not fun to debug missing shared libraries while running tfc executables. It is more straight forward to get the error while installing the package

exes/ethercat/CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ install(
   FILES
     systemd/ethercat.service
   DESTINATION
-    ${CMAKE_INSTALL_LIBDIR}/systemd/user/
+    /usr/lib/systemd/system/
```
jbbjarnason, line 50: why does ${CMAKE_INSTALL_LIBDIR} not work?

omarhogni: Sometimes it is the wrong thing see.  https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html 

And it kind of does not make sense. Why try to extract a specific location from variables.
This placement is not variable.

.github/workflows/compile.yml
omarhogni: This is supposed to upload information to github to show deps. https://devblogs.microsoft.com/cppblog/vcpkg-integration-with-the-github-dependency-graph/

.github/workflows/compile.yml
omarhogni: This is related to the dependency graph

.github/workflows/ubuntu-packaging.yml
omarhogni: That kind of depends on how we design our release structure, I was actually sitting on a local commit that restricts this to relases. And then my thougt was we would create a tag.

cmake/tfc_cpack.cmake
omarhogni: libc6 is not specific enough we need to then have the version also, cockpit is not a missing shared library, you can run all of tfc perfectly fine without it.

.github/workflows/compile.yml
jbbjarnason: Awesome 👍

.github/workflows/ubuntu-packaging.yml
jbbjarnason: For now I would say main is sufficient, but tags would be even better, we can obviously start tagging main with small incremental releases

cmake/tfc_cpack.cmake
jbbjarnason: Yes remove cockpit and use libc6 and even better add libstdc++ with their according version. For now maybe only remove cockpit?

exes/ethercat/CMakeLists.txt
jbbjarnason: I think you will break vcpkg usage of the framework. So I would like to investigate or test the vcpkg. This can also be postponed to a separate issue ticket.

omarhogni: Let's postpone this until confirmed to be an issue. Can be built using BUILD_EXES=False for now.

.github/workflows/ubuntu-packaging.yml
omarhogni: Let's do that. I will create a tagged inital release after this is merged as 2023.10.1

jbbjarnason (approved):

cmake/tfc_cpack.cmake
omarhogni: libc6 added back with version

PR #151, opened 2023-09-08
Adding SSL and no SSL option to MQTT broadcaster

magni-mar:


omarhogni:
Can this be closed in favor of #153

PR #155, opened 2023-09-14
Ipc tinker doc

omarhogni:


SindriTh (approved):

docs/functionality/ipc_tinker.md
```diff
@@ -0,0 +1,45 @@
...
+## General information
+The slot requests a name from dbus `<slot_name>._slot_`
+storing its dbus connection, 
+The implemented functionality inject their interfaces
+into the slots established dbus connection.
+
+### Interface names
+the following interface names are used for the functionality.
+
+- filters `<slot_name>/filters`
```
jbbjarnason, line 15: This might be prone to change but I will see how two bypass "config" in interface name within confman

docs/functionality/ipc_tinker.md
```diff
@@ -0,0 +1,45 @@
...
+- value `<slot_name>/value`
+
+## Specification of the tinker interface
+property called `forced` of the type bool,
+this property indicates weather the slots value is forced or not.
+
+a method called `force` accepting a value of the type of the slot,
+this method forces the value of the slot.
+
+a method called `release` accepting no arguments.
```
jbbjarnason, line 26: What does `release` do?

docs/functionality/ipc_tinker.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Specification of the value interface
+property called `post_filter_value` of the type of the slot.
```
jbbjarnason, line 29: Begin sentence with capital letter

jbbjarnason, line 29: `post_filter_value` or `filtered_value`

jbbjarnason (approved):

docs/functionality/ipc_tinker.md
omarhogni: release is unforce or releasing the force of the value.

omarhogni: Specified in doc.

docs/functionality/ipc_tinker.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Specification of the tinker interface
+A property called `forced` of the type bool,
+this property indicates weather the slots value is forced or not.
+
+A method called `force` accepting a value of the type of the slot,
+this method forces the value of the slot.
+
+A method called `release` accepting no arguments.
+Which releases the force on the slot having
```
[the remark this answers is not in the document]
omarhogni: The idea is to "stick" a value.
So no, the next value should not override it.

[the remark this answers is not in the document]
jbbjarnason: I think it would be safer to only implement a setter, you can set the value of the slot but it won't affect production as the next event will still be functional. I think it could be more harmful than good to block values from the signal.

So the use cases I have in mind for this feature is:
- IO testing inhouse manufacturing
- Service debugging 
- Optimizing pneumatic actuator timings
- Stop and start conveyors
- Change speed temporarily of a conveyor

I don't think any of those use cases require complete control of slot.

Do you have use cases in mind which would require blocking the following signal values?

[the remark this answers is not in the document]
omarhogni: So this idea is modeled after
the plc world.

But, there we have a cyclic logic loop so many variables are overwritten in each cycle.

Here we only write on change so a setter is propably sufficient.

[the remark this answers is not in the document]
omarhogni: @jbbjarnason I have updated the text to reflect this change.

jbbjarnason (approved):

PR #161, opened 2023-09-20
Reduce binary size

jbbjarnason:
Tried and failed/deferred, let's get more experience with completion_token support. For async_send&async_receive


jbbjarnason:
let's wait for modules support in cmake

PR #162, opened 2023-09-25
add stdlib to linker flags for llvadd stdlib to linker flags for llvmm

jbbjarnason:


.github/workflows/compile.yml
omarhogni, line 1: I really like this change. Simplifies this.

omarhogni (approved):
LGTM

PR #163, opened 2023-09-26
Time point not loosing precision in json conversion

jbbjarnason:
Explicit millisecond clock made with the iso8601 conversion.

Add uuid to item


libs/ipc/src/item.cpp
```diff
@@ -0,0 +1,38 @@
...
+  static std::mt19937_64 random_engine{};  // NOSONAR
+  static bool first_call{ true };
+  if (first_call) {
+    first_call = false;
+    auto proc_id{ std::hash<std::string>{}(fmt::format("{}{}", tfc::base::get_exe_name(), tfc::base::get_proc_name())) };
+    random_engine.seed(proc_id);
+  }
+  uuids::basic_uuid_random_generator<std::mt19937_64> random_generator{ random_engine };
+
+  return { .item_id = random_generator(), .entry_timestamp = std::chrono::system_clock::now() };
```
omarhogni, line 23: Having the state here global makes testing hard and the function unnecessarily complex.
I would like to see the global state removed from the basic make function
and rather a seed accepted. The global state can then be stored somewhere
else if required but the basic function pure.

Another thing is that I would like us to consider PCG rng instead of the mt
just because of size and initialization time.

libs/stx/inc/public/tfc/stx/glaze_meta.hpp
```diff
@@ -53,6 +52,23 @@ struct glz::meta<std::chrono::duration<rep_t, period_t>> {
...
+};
+template <typename clock_t, typename duration_t>
+struct glz::meta<std::chrono::time_point<clock_t, duration_t>> {
+  static constexpr std::string_view prefix{ "std::chrono::time_point<" };
+  static constexpr std::string_view postfix{ ">" };
+  static constexpr std::string_view separator{ "," };
+  static constexpr auto name{
+    tfc::stx::string_view_join_v<prefix, glz::name_v<clock_t, true>, separator, glz::name_v<duration_t, true>, postfix>
+  };
+};
```
omarhogni, line 71: This creates a non standard time representation that requires custom parsing in each platform
wanting to read it.
I don't see a use case for this level of precision being sent between systems.

as a milliseconds since 1 Jan 1970 I would not be opposed or simply seconds since 1970
this also loses all timezone information.

What does the json from this look like? is it just a number or is the duration period encoded as well?

does the receiving party need the json schema to realize the resolution from parsing the name?

vcpkg.json
```diff
@@ -17,11 +17,11 @@
     "boost-asio",
     "boost-program-options",
     "boost-stacktrace",
-    "date",
```
omarhogni, line 20: I think we can leave this in, it is tiny and really useful when working with dates.

libs/stx/inc/public/tfc/stx/glaze_meta.hpp
```diff
@@ -64,38 +80,6 @@ struct glz::meta<tfc::stx::basic_fixed_string<char_type, len>> {
...
-template <>
-struct to_json<std::chrono::time_point<std::chrono::system_clock>> {
-  template <auto Opts>
-  static void op(auto& value, auto&&... args) noexcept {
-    std::stringstream str_stream;
-    str_stream << date::format("%FT%T%Ez", value);
-    write<json>::op<Opts>(str_stream.str(), args...);
-  }
-};
-
```
omarhogni, line 98: Instead of removing this functionality could we find a way to be able to choose?
maybe create a wrapper type that switches the encoding of the date to a string 
format

omarhogni (changes requested):

libs/ipc/testing/tests/item_test.cpp
```diff
@@ -4,17 +4,32 @@
 #include <boost/ut.hpp>
```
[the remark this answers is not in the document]
jbbjarnason: not sure what you are referring to here?

libs/ipc/src/item.cpp
jbbjarnason: now using pcg as proposed

libs/stx/inc/public/tfc/stx/glaze_meta.hpp
jbbjarnason: > This creates a non standard time representation that requires custom parsing in each platform
wanting to read it.
I don't see a use case for this level of precision being sent between systems.

Now time_point of milliseconds is represented in iso8601 format.

> as a milliseconds since 1 Jan 1970 I would not be opposed or simply seconds since 1970
this also loses all timezone information.

timezone info is dependent on system_clock so this would not work robustly across different hosts

> What does the json from this look like? is it just a number or is the duration period encoded as well?

Time_point in micro or nanoseconds are represented as number, resulting in no truncation.

does the receiving party need the json schema to realize the resolution from parsing the name?

Ideally the receiving party knows the type

libs/ipc/testing/tests/item_test.cpp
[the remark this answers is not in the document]
omarhogni: There were compile errors. Missing reference to the item::make. Seems resolved.

libs/stx/inc/public/tfc/stx/glaze_meta.hpp
omarhogni: This just looks like something we don't want to be doing. What use case are we solving with this?

jbbjarnason: removed

omarhogni (approved):
Okay, I can agree to this.
We should add a static assert for nanoseconds
and microseconds if we don't want
them encoded.

jbbjarnason:
> Okay, I can agree to this. We should add a static assert for nanoseconds and microseconds if we don't want them encoded.

the `requires` clause takes care of nano and microseconds, `requires(!std::is_same_v<duration_t, std::chrono::microseconds> && !std::is_same_v<duration_t, std::chrono::nanoseconds>)` so when/if we need that precision we can review this decision.

PR #164, opened 2023-09-28
Button

omarhogni:
Added a exe to transform a bool signal into some events for a button.

### DIFF
--- a/exes/button/CMakeLists.txt
+++ b/exes/button/CMakeLists.txt
@@ -1,0 +1,38 @@
+add_executable(button src/main.cpp)
+
+find_path(AZMQ_INCLUDE_DIRS "azmq/actor.hpp")
+
+target_include_directories(button
+  PUBLIC
+    ${AZMQ_INCLUDE_DIRS}
+)
+
+find_package(Boost REQUIRED COMPONENTS program_options)
+find_package(glaze CONFIG REQUIRED)
+
+target_link_libraries(button
+  PUBLIC
+    tfc::ipc
+    tfc::base
+    tfc::logger
+    tfc::confman
+    Boost::program_options
+    glaze::glaze
+)
+
+include(GNUInstallDirs)
+install(
+  TARGETS
+    button
+  DESTINATION
+    ${CMAKE_INSTALL_BINDIR}
+  CONFIGURATIONS Release
+)
+
+install(
+  TARGETS
+    button
+  DESTINATION
+    ${CMAKE_INSTALL_BINDIR}/debug/
+  CONFIGURATIONS Debug
+)
--- a/exes/button/CMakeLists.txt
+++ b/exes/button/CMakeLists.txt
@@ -1,13 +1,8 @@
 add_executable(button src/main.cpp)
 
-find_path(AZMQ_INCLUDE_DIRS "azmq/actor.hpp")
+find_path(BEXT_SML_INCLUDE_DIRS "boost/sml.hpp")
+target_include_directories(button PRIVATE ${BEXT_SML_INCLUDE_DIRS})
 
-target_include_directories(button
-  PUBLIC
-    ${AZMQ_INCLUDE_DIRS}
-)
-
-find_package(Boost REQUIRED COMPONENTS program_options)
 find_package(glaze CONFIG REQUIRED)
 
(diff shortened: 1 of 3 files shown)


exes/button/src/main.cpp
```diff
@@ -0,0 +1,153 @@
...
+struct my_deps {
+  tfc::ipc::bool_signal& long_signal;
+  tfc::ipc::bool_signal& double_signal;
+  tfc::ipc::bool_signal& touch_signal;
+  asio::io_context& ctx;
+  tfc::logger::logger& logger;
+  std::chrono::time_point<system_clock, std::chrono::nanoseconds> initial_time;
+};
+
+// guards
```
jbbjarnason, line 36: lambdas can be constexpr

exes/button/src/main.cpp
```diff
@@ -0,0 +1,153 @@
...
+struct high {};
+struct low {};
+struct timeout {};
+
+struct my_deps {
+  tfc::ipc::bool_signal& long_signal;
+  tfc::ipc::bool_signal& double_signal;
+  tfc::ipc::bool_signal& touch_signal;
+  asio::io_context& ctx;
+  tfc::logger::logger& logger;
```
jbbjarnason, line 32: how about owning these references?

omarhogni: Sure, that looks better. ctx and logger kept as a reference.

exes/button/src/main.cpp
omarhogni: I will make them constexpr. Is there anything specific you hope to gain in this case?

jbbjarnason: not being globally constructed/destructed, instead being part of the compiled artifact(a symbol). I thought this hsould provoke an clang warning

exes/button/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+find_package(glaze CONFIG REQUIRED)
+
+target_link_libraries(button
+  PUBLIC
+    tfc::ipc
+    tfc::base
+    tfc::logger
+    tfc::confman
+    Boost::program_options
+    glaze::glaze
```
jbbjarnason, line 20:
```suggestion
```

exes/button/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+add_executable(button src/main.cpp)
+
+find_path(AZMQ_INCLUDE_DIRS "azmq/actor.hpp")
```
jbbjarnason, line 3: azmq is not explicitly used, but sml is used, so maybe change to sml

exes/button/src/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+// Actions
+constexpr auto send_signal = [](asio::io_context& ctx, tfc::ipc::bool_signal& signal, tfc::logger::logger& logger) {
+  signal.async_send(true, [&](const std::error_code& err, size_t) {
+    if (err) {
+      logger.warn("Failed to send, {}", err.message());
+    }
+  });
+
+  // Create a timer from a shared_ptr
+  auto timer = std::make_shared<asio::system_timer>(ctx);
```
jbbjarnason, line 83: my system instead of steady timer?

exes/button/CMakeLists.txt
omarhogni: copy paste error... fixed

exes/button/src/main.cpp
omarhogni: My thought was that steady timer is not wall time. But we are interacting with the physical world ( someones finger ) but with such large timers I don't think we will find any difference.

The only guarantee steady timer gives is that every call to the timer will return a later point then the last.

omarhogni: But then again if the computer applies time corrections during the time I am not sure what the behavior is.

jbbjarnason: I think it makes sense to use steady timer for deadline timers. System timer for absolute timers, for example 24.december at 12

omarhogni: I think you are correct, it looks like steady_timer is a rename of an older concept called `deadline_timer`. Will fix.

PR #166, opened 2023-09-28
Trying to make create_link more robust

### LINKED ISSUE #165: If .local cockpit folder does not exist the script fails loudly
https://github.com/Skaginn3x/framework/blob/729586370b88364130eea8d3510fbc0af5bc4826/cockpit/IPC/install/create_link.js#L6

Maybe create it instead.

SindriTh:
Fixes #165

### DIFF
--- a/cockpit/IPC/install/create_link.js
+++ b/cockpit/IPC/install/create_link.js
@@ -7,8 +7,8 @@ const targetDir = path.join(os.homedir(), '.local', 'share', 'cockpit');
 
 // Ensure the target directory exists
 if (!fs.existsSync(targetDir)) {
-    console.log(`Target directory ${targetDir} does not exist. Creating...`);
-    fs.mkdirSync(targetDir, { recursive: true });
+  console.log(`Target directory ${targetDir} does not exist. Creating...`);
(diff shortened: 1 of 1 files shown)


cockpit/IPC/install/create_link.js
```diff
@@ -1,26 +1,42 @@
...
-const moduleFolderName = moduleRootDir.substring(moduleRootDir.lastIndexOf('/') + 1, moduleRootDir.length);
+const moduleFolderName = path.basename(moduleRootDir);
 console.log(`Linking ${targetDir} to ${moduleRootDir} as ${moduleFolderName}`);
 
+// Check if symlink already exists
+if (fs.existsSync(moduleFolderName)) {
+    console.log(`A file or directory with the name ${moduleFolderName} already exists in ${targetDir}.`);
+    process.exit(1);
+}
+
```
omarhogni, line 28: Does this need to be an error? this is run as part of npm run devbuild and could just be the second time somebody ran devbuild.

SindriTh: Correct, This should exit with 0

omarhogni (approved):
LGTM

PR #170, opened 2023-10-02
Hardening Operation mode 

jbbjarnason:
- Add mode getter in interface
- Populate stopped state when operation mode is started
- Add tests 
- Emergency can be now entered from any state
- Fixed enum concept in sdbusplus meta


exes/operations/inc/state_machine_owner.hpp
```diff
@@ -0,0 +1,129 @@
...
+template <typename owner_t>
+struct state_machine;
+
+struct storage {
+  std::optional<std::chrono::milliseconds> startup_time{};
+  std::optional<std::chrono::milliseconds> stopping_time{};
+  struct glaze {
+    // clang-format off
+    static constexpr auto value{ glz::object(
+        "startup_time", &storage::startup_time, "[ms] Delay to run initial sequences to get the equipment ready.",
```
omarhogni, line 33: Is [ms] required. The frontend converts and marks this type.

exes/operations/inc/details/state_machine_owner_impl.hpp
omarhogni, line 1: Why is this entire file not inside a class definition.
This is really complex and repetitive for what it does,
so much boiler plate.

libs/ipc/testing/mocks/inc/public/tfc/mocks/ipc.hpp
omarhogni, line 1: Does this code segment have to be updated in a operation mode overhaul?

omarhogni (changes requested):
Seems overly complicated. Saving the definitions on a exe level is something that
is hard to follow. And then you have implementation code inside a hpp file ( state_machine_owner).

The addition of the tests is a big plus.

exes/operations/inc/state_machine_owner.hpp
jbbjarnason: not really

exes/operations/inc/details/state_machine_owner_impl.hpp
jbbjarnason: The reason is to compile this code in one translation unit (for production build), so instead of exposing symbols for testing object and moving this to a source file, I moved it to a implementation file. Then I can expose symbols in the source file for this implementation for production level code, and for testing I can include this implementation file explicitly and use the templated code.

libs/ipc/testing/mocks/inc/public/tfc/mocks/ipc.hpp
jbbjarnason: It simplifies the code, it can be extracted to separate PR which this PR could depend on. But these are pretty minor changes.

jbbjarnason: It came already with this PR https://github.com/Skaginn3x/framework/pull/163

exes/operations/inc/state_machine_owner.hpp
jbbjarnason: removed

exes/operations/inc/details/state_machine_owner_impl.hpp
jbbjarnason: Not related to templating, some projects use `.ipp` for implementation header files, like boost asio.

omarhogni: But this is a private header inside a our exe folder. This is not a library impl that gets used over and over again.

jbbjarnason: If everything is in one translation unit, main.cpp it will slow down the compilation as it would only be able to use one core. So when compiling in this manner one core can compile the implementation, and some other core can compile main or other units.

PR #182, opened 2023-10-06
Fix vcpkg port

jbbjarnason:


ports/tfc-framework/portfile.cmake
```diff
@@ -14,6 +14,7 @@ endif()
 vcpkg_cmake_configure(
     SOURCE_PATH "${SOURCE_PATH}"
     OPTIONS
+      -DCMAKE_PROJECT_VERSION=${VERSION} # todo I don't like this
```
omarhogni, line 17: This is crazy, we should maybe move the version back to CMAkelists.txt and I can modify the version sanity checker to parse that file instead.

vcpkg-test/vcpkg.json
```diff
@@ -10,5 +10,11 @@
       "host": true
     },
     "tfc-framework"
+  ],
+  "overrides": [
+    {
+      "name": "mp-units",
+      "version": "0.8.0"
+    }
```
omarhogni, line 18: This is horrible.

mp-units has announced they will not support the 0.8.0 tree
after the release of 2.0 so we should try and update quickly.

versions/t-/tfc-framework.json
```diff
@@ -1,7 +1,7 @@
 {
   "versions": [
     {
-      "git-tree": "2436086545723f9d673b8b5b900902e2f5d87167",
+      "git-tree": "6b86673d223034a66c1646bb7f8b62ab1caa0c75",
```
omarhogni, line 4: Can we populate this automatically? With this hand picked it is not certain that
you are installing the code from you current commit.

PR #183, opened 2023-10-06
Automatic creation of systemctl files

magni-mar:


omarhogni:
I dont think this is worth it right now.  Let's revisit when we have a greater count of executables. Closing.

omarhogni (approved):

PR #185, opened 2023-10-06
Adding config file to Ethercat.

magni-mar:


jbbjarnason:
@magni-mar why was this closed? 

To populate interfaces to json schema during runtime you could define to_json_schema for the given type. 

Example of implementation:
```cpp
namespace tfc::ethernet {
struct interface{
std::string value{};
static std::vector<interface> const& get() {
// This function should be implemented in source file and would return a reference to a global symbol.
// That will allow making string_view in to_json_schema call to this refernce.
// grab interfaces from system and populate vector to global variable within the source file for each call
return {};
}
std::string operator() const {
return value;
}
};
} // namespace tfc::ethernet
namespace tfc::json::detail {

template <>
struct to_json_schema<tfc::ethernet::interface> {
  template <auto opts>
  static void op(auto& schema, auto& defs) {
    // std::span schema.enumeration to view of tfc::ethernet::interface::get() 
  }
};
}


```

PR #191, opened 2023-10-11
Base the tfc-toolchain on an older distro

omarhogni:


jbbjarnason:
You will probably need to add static libstdc++ link flag to packaging pipeline

jbbjarnason:
And add build step for cross compiler. Or use clang for packing

PR #197, opened 2023-10-16
On the spot el1809 driver

jbbjarnason:
todo refactor interpretation of buffer


omarhogni:
@jbbjarnason This has been IO tested on a 1809 and is working. Ok to merge?

PR #203, opened 2023-10-17
Migrate the project to use modules

omarhogni:


jbbjarnason:
Compilers which CMake natively supports module dependency scanning include:

MSVC toolset 14.34 and newer (provided with Visual Studio 17.4 and newer)

LLVM/Clang 16.0 and newer

GCC 14 (for the in-development branch, after 2023-09-20) and newer

We need to update to gcc14

omarhogni:
> We need to update to gcc14

gcc 14 rumored to be released in the early months of 2024, https://gcc.gnu.org/gcc-14/changes.html

jbbjarnason:
> > We need to update to gcc14
> 
> gcc 14 rumored to be released in the early months of 2024, https://gcc.gnu.org/gcc-14/changes.html

Let's build it and modify the toolchain files accordingly, no need to wait.

omarhogni:
Closing for now. This is diverging from main quickly and there are still quite a few barriers to doing this properly.

PR #209, opened 2023-10-18
fixing operation mode

jbbjarnason:


libs/operation_mode/src/operation_mode.cpp
```diff
@@ -78,23 +77,27 @@ auto interface::operator=(interface&& to_be_erased) noexcept -> interface& {
...
+  constexpr auto make_transition_filter{ [](transition_e trans) noexcept {
     return [trans](callback_item const& itm) { return itm.transition == trans; };
   } };
-  static constexpr auto make_mode_filter{ [](mode_e mode) noexcept {
+  constexpr auto make_mode_filter{ [](mode_e mode) noexcept {
     return [mode](callback_item const& itm) { return itm.mode == mode; };
   } };
 
-  static const auto invoke{ [this, update_msg](callback_item const& itm) noexcept {
+  const auto invoke{ [this, update_msg](callback_item const& itm) noexcept {
```
jbbjarnason, line 100: i am guessing this was causing some issues

Issue #12, opened 2023-03-29
confman further development


jbbjarnason:
- Log configuration changes
- Make configuration management available on json-rpc ( cockpit f.e. )
- Services declare configuration fields, type, **limits** and defaults to configuration manager over ipc.

omarhogni:
Points 2 & 3 addressed already. Issue updated to reflect that.

jbbjarnason:
We could fix the disk writing problem in parallel. And maybe skip the part of logging, my suggestion would be to keep 10 copies of old config.

1. Write to file function would create file `<config_name>.json.tmp`
2. `rm` file 10
3. mv files 1-9 to 2-10 (`<config_name>.json.1) etc
4. cp file  `<config_name>.json` to  `<config_name>.json.1`
5. mv file `<config_name>.json.tmp` to `<config_name>.json`

The tmp file could be used as lock file, if we would like to detect failure in this process.

Lastly 10 could be default but this could be an argument or change-able parameter in confman API.

For frontend, populate a list of timestamp of those 10 files. Frontend can call a method with argument integer with indices of the timestamp to fetch and compare in UI.

omarhogni:
I think this is allot of work.
Can we wait with removing config and when writing a new config just write the old as well timestamped?

omarhogni:
Idea proposed in #218

omarhogni:
@magni-mar Can you implement the configuration retention?

magni-mar:
I can start working on this issue, can not promise any deadline though as I have no idea how long this will take.

PR #219, opened 2023-10-20
[libconfman] Retention policy

magni-mar:
See #12 And [policy](https://skaginn3x.github.io/framework/design/cm.html#configuration-retention-policy)


libs/confman/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ target_include_directories(confman
 target_link_libraries(confman
   PUBLIC
     Boost::boost
+    Boost::uuid
```
jbbjarnason, line 28: use similar uuid as in item.hpp

magni-mar: tried to mimick item.hpp, please see changes.

jbbjarnason:
This feature can be done behind the scenes, in other words in a source file. 
And please make use of STD:: filesystem as most as you can instead of manual string manipulation

libs/confman/testing/tests/file_storage_test.cpp
```diff
@@ -134,5 +141,34 @@ auto main(int argc, char** argv) -> int {
...
+    bool backup_found = false;
+    std::string found_file;
+
+    for (const auto& file : files) {
+      if (file.path().string().find("/tmp/test_uuid_") != std::string::npos) {
+        backup_found = true;
+        found_file = file.path().string();
+        break;
+      }
+    }
```
omarhogni, line 159: please use https://en.cppreference.com/w/cpp/filesystem/exists

libs/confman/testing/tests/file_storage_test.cpp
omarhogni, line 1: There are no cases here testing the functionality or policy. Please add them.

libs/confman/inc/public/tfc/confman/file_storage.hpp
```diff
@@ -12,6 +15,9 @@
 
 namespace tfc::confman {
 
+auto write_to_file(const std::filesystem::path& file_path, std::string_view file_contents) -> std::error_code;
+auto write_and_delete_old_files(std::string_view file_content, std::filesystem::path const& file_path) -> std::error_code;
```
omarhogni, line 19: I really don't like the name of this function. Does it need to exist?
Can the users not just 
```
write_to_file();
apply_retention_policy():
```
The delete old files is really a lie. These are not necessarily old files.

libs/confman/inc/public/tfc/confman/file_storage.hpp
```diff
@@ -83,22 +89,29 @@ class file_storage {
...
   auto set_changed() const noexcept -> std::error_code {
-    std::string buffer{};  // this can throw, meaning memory error
-    glz::write<glz::opts{ .prettify = true }>(storage_, buffer);
-    auto glz_err{ glz::buffer_to_file(buffer, config_file_.string()) };
-    if (glz_err != glz::error_code::none) {
-      logger_.warn(R"(Error: "{}" writing to file: "{}")", glz::write_json(glz_err), config_file_.string());
-      return std::make_error_code(std::errc::io_error);
-      // todo implicitly convert glaze error_code to std::error_code
+    std::string buffer{ to_json() };  // this can throw, meaning memory error
+    if (auto write_error{ write_to_file(config_file_, buffer) }; write_error) {
```
omarhogni, line 101:
```suggestion
    if (auto write_error{ write_to_file(config_file_, to_json()) }; write_error) {
```

libs/confman/inc/public/tfc/confman/file_storage.hpp
```diff
@@ -1,6 +1,9 @@
 #pragma once
+
 #include <sys/inotify.h>
+#include <chrono>
```
omarhogni, line 4: This can be included in the cpp file.

libs/confman/src/file_storage.cpp
```diff
@@ -0,0 +1,141 @@
...
+/// get_environment_variable("MAX_CONNECTIONS", 5); If the environment variable "MAX_CONNECTIONS" is set to "10", the
+/// function returns 10. Otherwise, the function returns 5.
+static auto get_environment_variable(std::string const& env_variable, int default_value) -> int {
+  if (const char* value = std::getenv(env_variable.c_str()); value != nullptr) {
+    if (int const value_i = std::stoi(value); value_i >= 0) {
+      return value_i;
+    }
+  }
+  return default_value;
+}
```
omarhogni, line 38: You should rename this or split into two. The "hidden" conversion to an int based on the name is not great.

Why is value_i of 0 invalid?
maybe you would like to store 0 DAYS but always 10 copies.

std::stoi can throw. Let's catch that and log rather then crashing.
see https://en.cppreference.com/w/cpp/string/basic_string/stol

libs/confman/src/file_storage.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+/// \brief Fetches the value of the environment variable "TFC_CONFMAN_MIN_RETENTION_COUNT", default value is 4.
+/// \return minimum retention count set in the environment variable
+/// \example get_minimum_retention_count();
+/// If the environment variable "TFC_CONFMAN_MIN_RETENTION_COUNT" is set to "6", the function returns 6.
+/// Otherwise, the function returns 4.
+static auto get_minimum_retention_count() -> int {
+  return get_environment_variable("TFC_CONFMAN_MIN_RETENTION_COUNT", 4);
+}
```
omarhogni, line 56: I would like to see size_t used as a signature. The value should not 
be negative. What does the user logic do with a negative value?

libs/confman/src/file_storage.cpp
```diff
@@ -0,0 +1,141 @@
...
+  uint64_t count = 0;
+  for (auto it = file_times.begin();
+       it != file_times.end() && count <= (file_times.size() - static_cast<uint64_t>(total_file_retention_count));
+       ++it, ++count) {
+    auto time_since_last_modified = current_time - it->first;
+    if (time_since_last_modified > retention_time) {
+      std::filesystem::remove(it->second);
+    }
+  }
+}
```
omarhogni, line 91: Please split this function up so you can unit test the user logic.
If the functions have side effects you make testing harder.

You can write "pure" functions and then string them together in a top level function.

omarhogni (changes requested):

libs/confman/testing/tests/file_storage_test.cpp
magni-mar: Since the full path of the file is not known it's not possible to use `std::filesystem::exists`. A uuid is concatenated on the back of `/tmp/test_uuid_` which is random.

libs/confman/inc/public/tfc/confman/file_storage.hpp
magni-mar: I agree that the name should rather be `apply_retention_policy` but I disagree that the function should not be used. If this function is skipped then the header file needs to be exposed to more functions instead of just these two, in my opinion it is cleaner to only expose these 2.

libs/confman/src/file_storage.cpp
magni-mar: `value_i` can be 0, `value_i >= 0` means equal to or greater than 0.

libs/confman/src/file_storage.cpp
magni-mar: changed this to unsigned int.

libs/confman/testing/tests/file_storage_test.cpp
magni-mar: see new tests

libs/confman/src/file_storage.cpp
magni-mar: split the function and added more extensive testing, please see changes.

libs/confman/src/file_storage.cpp
```diff
@@ -0,0 +1,159 @@
...
+  }
+  return file_times;
+}
+
+/// \brief removes files that exceed the specified retention policy
+/// \param file_times map containing files sorted by their modification time
+/// \param retention_count maximum number of files to retain
+/// \param retention_time maximum age of files to retain
+/// \note in order for a file to be removed, it both needs to be too old and too many files
+void remove_files_exceeding_retention(const std::map<std::filesystem::file_time_type, std::filesystem::path>& file_times,
```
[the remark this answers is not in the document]
magni-mar: moved functions only used within the translation unit to anonymous namespace and then a function that is used in the translation unit and the testing file to a header file, both the tests and the translation unit then import the header file.

libs/confman/src/file_storage.cpp
```diff
@@ -0,0 +1,159 @@
+#include <chrono>
+#include <cmath>
+#include <cstdint>
+#include <cstdlib>
+#include <iostream>
```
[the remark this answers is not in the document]
magni-mar: nothing, just an accidental include

libs/confman/src/file_storage.cpp
```diff
@@ -0,0 +1,159 @@
+#include <chrono>
+#include <cmath>
+#include <cstdint>
+#include <cstdlib>
```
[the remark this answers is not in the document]
magni-mar: thanks :)

libs/confman/src/file_storage.cpp
```diff
@@ -0,0 +1,159 @@
...
+  } catch (std::invalid_argument const&) {
+    return 4;
+  } catch (std::out_of_range const&) {
+    return 4;
+  }
+}
+
+/// \brief get total file count in a directory
+/// \param directory path to directory
+static auto get_total_file_count(std::filesystem::path const& directory) -> size_t {
```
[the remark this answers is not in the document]
magni-mar: implemented

libs/confman/src/file_storage.cpp
```diff
@@ -0,0 +1,159 @@
...
+  return std::distance(std::filesystem::directory_iterator(directory), std::filesystem::directory_iterator{});
+}
+
+/// \brief get map of files in directory, sorting them by time
+/// \param directory path to directory
+static auto get_sorted_file_list(std::filesystem::path const& directory)
```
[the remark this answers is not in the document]
magni-mar: a map is sorted based on the key, in this case the time, so when iterating over the map they are iterated over in the order in which `last_write_time` is set.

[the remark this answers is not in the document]
magni-mar: the name can be confusing though, changing to `get_files_by_last_write_time`

libs/confman/src/file_storage.cpp
```diff
@@ -0,0 +1,159 @@
...
+static auto apply_retention_policy(std::filesystem::path const& directory) -> void {
+  // +2 for the file itself and the swap file
+  size_t const retention_count = get_minimum_retention_count() + 2;
+  std::chrono::days const retention_time = get_minimum_retention_days();
+
+  if (get_total_file_count(directory) <= retention_count) {
+    return;
+  }
+
+  std::map<std::filesystem::file_time_type, std::filesystem::path> const file_times = get_sorted_file_list(directory);
```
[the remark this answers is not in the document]
magni-mar: The iterator doesn't provide the files ordered by timestamp.

libs/confman/src/file_storage.cpp
```diff
@@ -0,0 +1,159 @@
...
+      std::filesystem::remove(path);
+    }
+  }
+}
+
+/// \brief Delete old files from the specified directory based on retention count and days. If the number
+/// of files exceed the retention count AND if the file's last modification date surpasses the retention time,
+/// \param directory path to directory containing files to possibly delete.
+static auto apply_retention_policy(std::filesystem::path const& directory) -> void {
+  // +2 for the file itself and the swap file
```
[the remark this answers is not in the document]
jbbjarnason: do you have an output from `ls` showing that file?

libs/confman/testing/tests/file_storage_test.cpp
```diff
@@ -134,5 +158,118 @@ auto main(int argc, char** argv) -> int {
...
+    ut::expect(std::distance(std::filesystem::directory_iterator(parent_path), std::filesystem::directory_iterator{}) == 2);
+
+    std::filesystem::remove_all(parent_path);
+  };
+
+  "write to file"_test = [&] {
+    std::filesystem::path const file_path{ std::filesystem::temp_directory_path() / "glaze file" };
+    std::string const file_content = R"(file content: {"username": "me"})";
+    tfc::confman::write_to_file(file_path, file_content);
+    ut::expect(std::filesystem::exists(file_path));
```
jbbjarnason, line 261: missing read_file expection

libs/confman/testing/tests/file_storage_test.cpp
```diff
@@ -134,5 +158,118 @@ auto main(int argc, char** argv) -> int {
...
+    }
+
+    ut::expect(backup_found) << "No backup file found for modified config";
+
+    if (backup_found) {
+      glz::json_t backup_json{};
+      std::string buffer{};
+      glz::read_file_json(backup_json, found_file.string(), buffer);
+
+      ut::expect((backup_json["a"].get<double>() - 3) < 0.01) << backup_json["a"].get<double>();
```
jbbjarnason, line 188: `a` is an integer why not read as integer?

libs/confman/src/file_storage.cpp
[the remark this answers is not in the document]
omarhogni: This is a vim artifact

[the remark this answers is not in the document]
omarhogni: Need some better semantics to skip over files. There could be any number of things inside that directory that are non of your concern.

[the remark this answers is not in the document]
jbbjarnason: Use std::views:: filter on file extension when creating map?

See example lib/operation mode source file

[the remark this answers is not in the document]
magni-mar: added a filter that filters on file ending, requiring `.json`. 

I think it's safe to assume that the only `.json` files that are in the directory are related to configuration

libs/confman/testing/tests/file_storage_test.cpp
magni-mar: Because `json_t` doesn't store `int`, it stores a variant `std::variant<null_t, double, std::string, bool, array_t, object_t>`

https://github.com/stephenberry/glaze/blob/main/include/glaze/json/json_t.hpp

libs/confman/src/file_storage.cpp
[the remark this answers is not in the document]
omarhogni: > added a filter that filters on file ending, requiring `.json`.
> 
> I think it's safe to assume that the only `.json` files that are in the directory are related to configuration

I dont agree with this
Screenshot from 2023-10-27 13-59-36
Each folder can have many files. You need to compare everything except for the uuid.

libs/confman/inc/public/tfc/confman/detail/retention.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+static constexpr std::string_view json_file_ending{ ".json" };
+
+/// \brief get map of files in directory, sorting them by time
+/// \param directory path to directory
+static auto get_json_files_by_last_write_time(std::filesystem::path const& directory)
+    -> std::map<std::filesystem::file_time_type, std::filesystem::path> {
+  std::map<std::filesystem::file_time_type, std::filesystem::path> file_times;
+
+  constexpr auto make_json_filter = [](std::filesystem::directory_entry const entry) noexcept {
```
jbbjarnason:
```suggestion
  constexpr auto make_json_filter = [](std::filesystem::directory_entry const& entry) noexcept {
```

libs/confman/src/file_storage.cpp
```diff
@@ -0,0 +1,123 @@
...
+/// \returns A std::error_code indicating success or failure. If write fails, IO error is returned.
+auto write_to_file(std::filesystem::path const& file_path, std::string_view file_contents) -> std::error_code {
+  auto glz_err{ glz::buffer_to_file(file_contents, file_path) };
+  if (glz_err != glz::error_code::none) {
+    return std::make_error_code(std::errc::io_error);
+  }
+  return {};
+}
+
+}  // namespace tfc::confman
```
jbbjarnason, line 123: missing new line?

libs/confman/src/file_storage.cpp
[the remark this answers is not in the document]
magni-mar: good point, added a more extensive filter and tests to test that filter, please see changes.

libs/confman/inc/public/tfc/confman/detail/retention.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include <string>
+#include <string_view>
+
+namespace tfc::confman {
+
+static constexpr std::string_view json_file_ending{ ".json" };
+
+/// \brief get map of files in directory, sorting them by time
+/// \param directory path to directory
+static auto get_json_files_by_last_write_time(std::filesystem::path const& directory)
```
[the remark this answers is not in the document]
magni-mar: the compiler complains if I do this

```bash
 error: no previous prototype for function 'get_json_files_by_last_write_time' [-Werror,-Wmissing-prototypes]
   21 | auto get_json_files_by_last_write_time(std::filesystem::path const& file_path)
      |      ^
```

I can add this to source file, add `inline`, or static, as far as I know

libs/confman/src/file_storage.cpp
magni-mar: no?

libs/confman/inc/public/tfc/confman/detail/retention.hpp
```diff
@@ -16,18 +16,21 @@ namespace tfc::confman {
...
     -> std::map<std::filesystem::file_time_type, std::filesystem::path> {
   std::map<std::filesystem::file_time_type, std::filesystem::path> file_times;
 
-  constexpr auto make_json_filter = [](std::filesystem::directory_entry const& entry) noexcept {
-    return entry.path().extension() == json_file_ending;
+  auto make_filter = [file_path](std::filesystem::directory_entry entry) noexcept {
+    /// get all files that have .json ending and have same start letters
+    /// example: test_file.json, should return true on test_file_1.json and test_file_2.json, but not test.json
+    const auto& stem = file_path.stem().string();
+    return entry.path().extension() == json_file_ending && entry.path().stem().string().substr(0, stem.size()) == stem;
```
jbbjarnason, line 29: how about something like `entry.path().filename().string().starts_with(file_path.filename().string());`

libs/confman/inc/public/tfc/confman/detail/retention.hpp
```diff
@@ -16,18 +16,21 @@ namespace tfc::confman {
...
-static auto get_json_files_by_last_write_time(std::filesystem::path const& directory)
+/// \brief get map of specified configuration in directory, sorting them by time
+/// \param file_path configuration file to get backups from
+static auto get_json_files_by_last_write_time(std::filesystem::path const& file_path)
     -> std::map<std::filesystem::file_time_type, std::filesystem::path> {
   std::map<std::filesystem::file_time_type, std::filesystem::path> file_times;
 
-  constexpr auto make_json_filter = [](std::filesystem::directory_entry const& entry) noexcept {
-    return entry.path().extension() == json_file_ending;
+  auto make_filter = [file_path](std::filesystem::directory_entry entry) noexcept {
```
jbbjarnason, line 25: file_path by reference capture

libs/confman/inc/public/tfc/confman/detail/retention.hpp
```diff
@@ -16,18 +16,21 @@ namespace tfc::confman {
...
+    /// get all files that have .json ending and have same start letters
+    /// example: test_file.json, should return true on test_file_1.json and test_file_2.json, but not test.json
+    const auto& stem = file_path.stem().string();
```
jbbjarnason, line 28: this should be a compiler error, you assign a rvalue to lvalue ref type

libs/confman/inc/public/tfc/confman/detail/retention.hpp
magni-mar: sounds good, but it needs to be `entry.path().filename().string().starts_with(file_path.stem().string())` to ignore the extension

libs/confman/inc/public/tfc/confman/detail/retention.hpp
magni-mar: this line is unnecessary anyways

libs/confman/inc/public/tfc/confman/detail/retention.hpp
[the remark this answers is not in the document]
magni-mar: changing to inline

libs/confman/inc/public/tfc/confman/detail/retention.hpp
```diff
@@ -0,0 +1,80 @@
...
+  }
+  return file_times;
+}
+
+/// \brief removes files that exceed the specified retention policy
+/// \param file_times map containing json files sorted by their modification time
+/// \param retention_count maximum number of files to retain
+/// \param retention_time maximum age of files to retain
+/// \note in order for a file to be removed, it both needs to be too old and too many files
+static auto remove_json_files_exceeding_retention(
```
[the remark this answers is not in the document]
magni-mar: changing to inline

jbbjarnason (approved):

omarhogni (approved):

PR #225, opened 2023-10-25
Dbus sml api

jbbjarnason:
Provide live view to running state machines generically.

Example, using https://github.com/seesleestak/plantuml-previewer-react:

With ref to @SindriTh [comment](https://github.com/Skaginn3x/framework/pull/225#issuecomment-1779401723) I have made available transitions from current state green, additionally, I have retained history being yellow, how you entered the current state.

image

image

Output from the plantuml is:

```plantuml
@startuml

[*] --> init
init -[#gold]-> stopped #limegreen : set_stopped / lambda
stopped : entry / lambda
stopped : exit / lambda
stopped -[#green]-> starting : set_starting / lambda
stopped -[#green]-> starting : run_button / lambda
starting : entry / lambda
starting : exit / lambda
starting --> running : starting_timeout / lambda
starting --> running : starting_finished / lambda
running : entry / lambda
running : exit / lambda
running --> stopping : run_button / lambda
running --> stopping : set_stopped / lambda
stopping : entry / lambda
stopping : exit / lambda
stopping --> stopped : stopping_timeout / lambda
stopping --> stopped : stopping_finished / lambda
stopped -[#green]-> cleaning : cleaning_button / lambda
stopped -[#green]-> cleaning : set_cleaning / lambda
cleaning : entry / lambda
cleaning : exit / lambda
cleaning --> stopped : cleaning_button / lambda
cleaning --> stopped : set_stopped / lambda
stopped -[#green]-> emergency : set_emergency / lambda
stopping --> emergency : set_emergency / lambda
starting --> emergency : set_emergency / lambda
running --> emergency : set_emergency / lambda
cleaning --> emergency : set_emergency / lambda
fault --> emergency : set_emergency / lambda
maintenance --> emergency : set_emergency / lambda
stopped -[#green]-> emergency : emergency_on / lambda
stopping --> emergency : emergency_on / lambda
starting --> emergency : emergency_on / lambda
running --> emergency : emergency_on / lambda
cleaning --> emergency : emergency_on / lambda
fault --> emergency : emergency_on / lambda
maintenance --> emergency : emergency_on / lambda
emergency : entry / lambda
emergency : exit / lambda
emergency --> stopped : emergency_off / lambda
stopped -[#green]-> fault : fault_on / lambda
stopped -[#green]-> fault : set_fault / lambda
running --> fault : fault_on / lambda
running --> fault : set_fault / lambda
fault : entry / lambda
fault : exit / lambda
fault --> stopped : fault_off / lambda
fault --> stopped : set_stopped / lambda
stopped -[#green]-> maintenance : maintenance_button / lambda
stopped -[#green]-> maintenance : set_maintenance / lambda
maintenance : entry / lambda
maintenance : exit / lambda
maintenance --> stopped : maintenance_button / lambda
maintenance --> stopped : set_stopped / lambda

@enduml



```


omarhogni:
This is a neat idea :+1:

jbbjarnason:
not ready, forgot to use constexpr name if it exists

SindriTh:
Is this something you want on the frontend? If so where would we put it?

jbbjarnason:
Or we could use https://arthursonzogni.github.io/FTXUI/ and display the statemachine with tfcctl

omarhogni:
I think that it would be nice to be able to do both honestly. But I am a little nervous about the amount of data we are transmitting over dbus when 99% of the time nobody is viewing. We will need to see how much this effects a running system. if at all.

SindriTh:
> I think that it would be nice to be able to do both honestly. But I am a little nervous about the amount of data we are transmitting over dbus when 99% of the time nobody is viewing. We will need to see how much this effects a running system. if at all.

We could do something like a heartbeat, where I toggle a parameter on DBus from the frontend every ~30 seconds and if you haven't heard from me in 60, then you can stop transmitting.
This would fail if the window is left open though.

SindriTh:
> > Is this something you want on the frontend? If so where would we put it?
> 
> I was thinking that it could be a separate tab, SM debugging
> 
> If we could pick running state machines and display their live statemachine in a dialog similarly to the watch IO debugging.
> 
> @omarhogni, @SindriTh thoughts on this suggestion?

I like this idea, Could we also show which parameters are needed to proceed to the next state? i.e a red or green circle showing if each requirement needed to proceed is fulfilled? 

This could enable us to see why a particular state machine isn't changing state.

omarhogni:
> We could do something like a heartbeat, where I toggle a parameter on DBus from the frontend every ~30 seconds and if you haven't heard from me in 60, then you can stop transmitting. This would fail if the window is left open though.

I dont think that is a good idea. Could have side affects activating the heartbeat. For now I think we keep this as a consideration and see if it causes any issues when the system gets some real runtime.

libs/dbus_util/inc/public/tfc/dbus/sml_interface.hpp
```diff
@@ -0,0 +1,219 @@
...
+  return boost::sml::aux::get_type_name<second_t>();
+}
+
+template <typename event_t>
+auto constexpr extract_event_type(event_t const& event) noexcept -> std::string {
+  if constexpr (tfc::stx::is_specialization_v<event_t, boost::sml::back::on_entry> ||
+                tfc::stx::is_specialization_v<event_t, boost::sml::back::on_exit>) {
+    return detail::extract_event_type(event);
+  } else {
+    return boost::sml::aux::get_type_name<event_t>();
```
omarhogni, line 55: I see, so you can override a name otherwise you get a sml provided name. That is pretty neat.

libs/dbus_util/inc/public/tfc/dbus/sml_interface.hpp
```diff
@@ -0,0 +1,219 @@
...
+  void log_action(action_t const& action, event_t const& event) {
+    logger::log_action<state_machine_t>(action, event);
+  }
+
+  template <class state_machine_t, class source_state_t, class destination_state_t>
+  void log_state_change(source_state_t const& src, destination_state_t const& dst) {
+    impl_.on_state_change(src.c_str(), dst.c_str(), last_event_);
+    std::stringstream iss{};
+    detail::dump<boost::sml::sm<state_machine_t>>(dst, iss);
+    impl_.dot_format(iss.str());
```
omarhogni, line 120: Can we do this without a stringstream?

libs/dbus_util/src/sml_interface.cpp
```diff
@@ -0,0 +1,56 @@
...
+namespace test {
+struct test_state {
+  static constexpr std::string_view name{ "test_state" };
+};
+struct invalid_test_state {
+  static constexpr std::string_view name1{ "test_state" };
+};
+static_assert(name_exists<test_state>);
+static_assert(!name_exists<invalid_test_state>);
+}  // namespace test
```
omarhogni, line 53: Do you want to have this here?

omarhogni (approved):

exes/operations/inc/details/state_machine.hpp
```diff
@@ -7,23 +7,58 @@
...
-struct emergency_off {};
-struct fault_on {};
-struct set_fault {};
-struct fault_off {};
-struct maintenance_button {};
-struct set_maintenance {};
+using std::string_view_literals::operator""sv;
+struct set_starting {
+  static constexpr auto name{ "set_starting"sv };
+};
```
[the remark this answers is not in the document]
jbbjarnason: `"my_event"_e` should display `"my_event"`, if name is not declared the default would be the whole typename `tfc::operation::detail::events::set_starting` which is rather bloated, so this offers a way to simplify the output, and of course not required at all.

libs/dbus_util/inc/public/tfc/dbus/sml_interface.hpp
jbbjarnason: Thank you, this is actually not doing the naming thing but it should! Will add.

libs/dbus_util/src/sml_interface.cpp
jbbjarnason: Since there is not a dedicated test file yes, and this should not affect in any way the produced binary.

libs/dbus_util/inc/public/tfc/dbus/sml_interface.hpp
jbbjarnason: yes, I just copied the example from boost sml which used stringstream, it should actually be quicker to use string https://stackoverflow.com/questions/19844858/c-stdostringstream-vs-stdstringappend so I will change to string

libs/dbus_util/inc/public/tfc/dbus/sml_interface.hpp
jbbjarnason: fixed here https://github.com/Skaginn3x/framework/pull/225/commits/51aa82e3e63a55f1ee621bc60cd81baeec6554ee

libs/dbus_util/inc/public/tfc/dbus/sml_interface.hpp
jbbjarnason: fixed here https://github.com/Skaginn3x/framework/pull/225/commits/b1b60c17de98847cd8073daa9b5a3abae14ad128

libs/dbus_util/inc/public/tfc/dbus/sml_interface.hpp
```diff
@@ -0,0 +1,219 @@
...
+/// int main() {
+///   boost::asio::io_context ctx{};
+///   auto bus = std::make_shared<sdbusplus::asio::connection>(ctx);
+///   auto interface = std::make_shared<sdbusplus::asio::dbus_interface>(bus, std::string{ tfc::dbus::sml::tags::path },
+///   "StateMachineName"); tfc::dbus::sml::interface sml_interface{ interface, "Log key" }; // optional log key
+///   // NOTE! interface struct requires to be passed by l-value like below, so the using code needs to store it like above
+///   boost::sml::sm<state_machine, boost::sml::logger<tfc::dbus::sml::interface>> sm{ sml_interface };
+///   return EXIT_SUCCESS;
+/// }
+/// \endcode
```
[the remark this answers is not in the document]
jbbjarnason: Now as well an example file, but I kept the copy of the code here so when hovering the class in IDEs you get the example in that hover dialog

PR #226, opened 2023-10-26
toolchain cmake v3.28

omarhogni:


omarhogni:
Missing dev build of gcc to be complete

jbbjarnason:
> Missing dev build of gcc to be complete

```bash
#!/usr/bin/env bash
set -ex

source shared.sh

GCC_VERSION=$1

curl https://ftp.gnu.org/gnu/gcc/gcc-$GCC_VERSION/gcc-$GCC_VERSION.tar.xz | xzcat | tar xf -
cd gcc-$GCC_VERSION

./contrib/download_prerequisites
mkdir ../gcc-build
cd ../gcc-build
hide_output ../gcc-$GCC_VERSION/configure \
    --prefix=/cpproot \
    --enable-languages=c,c++ \
    --disable-gnu-unique-object
hide_output make -j32
hide_output make install
ln -sf gcc /cpproot/bin/cc

cd ..
rm -rf gcc-build
rm -rf gcc-$GCC_VERSION
```

containers/tfc-toolchain.dockerfile
```diff
@@ -2,21 +2,36 @@ FROM ubuntu:23.04
 
 ENV DEBIAN_FRONTEND noninteractive
 
+WORKDIR /home/tfc/
+
 RUN apt-get update && apt-get install software-properties-common -y
+
+# Build development version of gcc
+RUN apt install build-essential libmpfr-dev libgmp3-dev libmpc-dev flex git -y
+RUN git clone https://github.com/gcc-mirror/gcc.git && cd gcc && ./configure --enable-languages=c,c++ --disable-multilib --prefix=/usr --program-suffix=-git && make -j$(nproc) && make install
```
jbbjarnason, line 11: add git checkout specific hash to know which "version" is actually available

omarhogni:
@jbbjarnason There are good changes in here with how we organize our toolchain. Can we back off with trying to support modules, build the current release versions of software and merge this?

jbbjarnason:
> @jbbjarnason There are good changes in here with how we organize our toolchain. Can we back off with trying to support modules, build the current release versions of software and merge this?

cross compiler is still missing, we can skip building gcc and use binaries from upstream and conclude this for now

omarhogni:
Current status is that all dependencies are built from source with the exception of gcc and aarch64-gcc which are installed from apt. As always more work here is needed. Would be nice to create a couple of containers ( documentation, gcc compile and clang compile) in order to pull in smaller container images.

cmake
jbbjarnason, line 1: Squashed cmake_modules to main, the new hash is: 9fb788699001a23065219752cdc5393a71dc5d4a

containers/tfc-toolchain/build-mold.sh
```diff
@@ -0,0 +1,20 @@
...
+hide_output cd $BUILD_DIR
+
+hide_output cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/cpproot/ ..
+hide_output cmake --build . -j $(nproc)
+hide_output cmake --build . --target install
+
+cd /tmp
+rm -rf mold-$MOLD_VERSION
+# Make mold the default linker
+#ln -sf /cpproot/bin/mold /usr/bin/ld
```
jbbjarnason, line 20:
```suggestion
```
I would suggest using toolchain files to make this switch

containers/tfc-toolchain/cpproot.conf
jbbjarnason, line 1: weird that this is not enough

jbbjarnason (approved):
looks good

PR #230, opened 2023-10-27
using last value cache from dbus instead of internal state

jbbjarnason:
should be faster and does not depend on async operations

### DIFF
--- a/libs/ipc/inc/public/tfc/ipc/details/dbus_ipc.hpp
+++ b/libs/ipc/inc/public/tfc/ipc/details/dbus_ipc.hpp
@@ -55,8 +55,7 @@ public:
   auto operator=(dbus_ipc&&) noexcept -> dbus_ipc& = default;
 
   void initialize() {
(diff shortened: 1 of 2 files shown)


libs/ipc/inc/public/tfc/ipc/details/dbus_ipc.hpp
```diff
@@ -44,30 +44,19 @@ class dbus_ipc {
...
-            if (auto current_value = value_getter_(); current_value.has_value()) {
-              return current_value.value();
-            }
-          } else {
-            return value_getter_();
-          }
-          return value_t{};
-        });
+    interface_->register_property_r<value_t>(std::string{ dbus::tags::value }, sdbusplus::vtable::property_::emits_change,
+                                             [](value_t& old_value) { return old_value; });
```
jbbjarnason, line 59: this is the fix

omarhogni (approved):

PR #237, opened 2023-10-31
Unnecessary if check

magni-mar:


exes/ethercat/inc/public/tfc/ec.hpp
```diff
@@ -233,14 +232,6 @@ class context_t {
...
     max_cycle_ = std::max(max_cycle_, last_cycle_);
 
-    if (cycle_count_ % 10'000 == 0 and false) {
-      // log the max cycle time
-      logger_.trace("Ethercat max cycle time: {} us", max_cycle_.count());
-      logger_.trace("Ethercat max cycle time with sleep: {} us", max_cycle_with_sleep_.count());
-      logger_.trace("Ethercat min cycle time: {} us", min_cycle_.count());
-      logger_.trace("Ethercat min cycle time with sleep: {} us", min_cycle_with_sleep_.count());
-    }
-
```
jbbjarnason, line 243: I would like to keep this for debugging purposes

jbbjarnason (changes requested):

exes/ethercat/inc/public/tfc/ec.hpp
```diff
@@ -181,10 +181,9 @@ class context_t {
             i + 1);
         async_wait(true);
         return {};
-      } else {
-        context_.slavelist[0].state = EC_STATE_OPERATIONAL;
-        ecx::write_state(&context_, 0);
```
omarhogni, line 186: This is being used, it is what happens while we wait for devices to reach operational.

exes/ethercat/inc/public/tfc/ec.hpp
jbbjarnason: dbus signals?

or just make it log, remove the `and false`

exes/ethercat/inc/public/tfc/ec.hpp
jbbjarnason: if stmt returns above

exes/ethercat/inc/public/tfc/ec.hpp
omarhogni: I was thinking ipc signals, but then again we are going to need some comprehensive way to debug the ethercat network. This could become one part of that and be exposed over dbus somehow.

jbbjarnason: why not use logging for debugging purposes?

exes/ethercat/inc/public/tfc/ec.hpp
omarhogni: It is in a loop. This path is taken often before the return is hit.

omarhogni: Many devices especially our vfd's take a long time to transition to OPERATIONAL ethercat state. approx 2.7seconds

PR #239, opened 2023-11-02
Motor proposal interface implementation

omarhogni:


omarhogni:
Closing, this PR went of course in trying to implement motor as modules. Harder then expected.

PR #247, opened 2023-11-06
atv320 ratio slot

jbbjarnason:
* Map function made constexpr and moved tests to being compiled
* Map caps input value according to input value mappings
* Atv320 change input freq to input ratio with -100% to 100% range

### DIFF
--- a/exes/ethercat/inc/public/tfc/cia/402.hpp
+++ b/exes/ethercat/inc/public/tfc/cia/402.hpp
@@ -28,9 +28,7 @@ struct control_word {
   bool reserved_4 : 1 {};
   bool reserved_5 : 1 {};
 
-  static constexpr auto from_uint(std::uint16_t word) noexcept -> control_word {
-    return std::bit_cast<control_word>(word);
-  }
+  static constexpr auto from_uint(std::uint16_t word) noexcept -> control_word { return std::bit_cast<control_word>(word); }
 
   constexpr explicit operator uint16_t() const noexcept {
     std::bitset<16> output{};
@@ -193,9 +191,7 @@ struct status_word {
     return states_e::not_ready_to_switch_on;
   }
 
-  static constexpr auto from_uint(uint16_t word) noexcept -> status_word {
-    return std::bit_cast<status_word>(word);
-  }
+  static constexpr auto from_uint(uint16_t word) noexcept -> status_word { return std::bit_cast<status_word>(word); }
 };
 
 static_assert(sizeof(status_word) == 2);
--- a/exes/ethercat/inc/public/tfc/cia/402.hpp
+++ b/exes/ethercat/inc/public/tfc/cia/402.hpp
@@ -29,22 +29,7 @@ struct control_word {
   bool reserved_5 : 1 {};
 
   static constexpr auto from_uint(std::uint16_t word) noexcept -> control_word {
-    return control_word{ .operating_state_switch_on = static_cast<bool>(word & 0b1),
-                         .enable_voltage = static_cast<bool>(word & 0b10),
-                         .operating_state_quick_stop = static_cast<bool>(word & 0b100),
-                         .enable_operation = static_cast<bool>(word & 0b1000),
-                         .operating_mode_specific_0 = static_cast<bool>(word & 0b10000),
-                         .operating_mode_specific_1 = static_cast<bool>(word & 0b100000),
-                         .operating_mode_specific_2 = static_cast<bool>(word & 0b1000000),
-                         .fault_reset = static_cast<bool>(word & 0b10000000),
-                         .halt = static_cast<bool>(word & 0b100000000),
-                         .operating_mode_specific_3 = static_cast<bool>(word & 0b1000000000),
-                         .reserved_0 = static_cast<bool>(word & 0b10000000000),
-                         .reserved_1 = static_cast<bool>(word & 0b100000000000),
-                         .reserved_2 = static_cast<bool>(word & 0b1000000000000),
-                         .reserved_3 = static_cast<bool>(word & 0b10000000000000),
-                         .reserved_4 = static_cast<bool>(word & 0b100000000000000),
-                         .reserved_5 = static_cast<bool>(word & 0b1000000000000000) };
+    return std::bit_cast<control_word>(word);
   }
 
   constexpr explicit operator uint16_t() const noexcept {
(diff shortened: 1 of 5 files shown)


exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320.hpp
```diff
@@ -552,8 +615,17 @@ class atv320 final : public base {
...
+      std::bitset<16> command_word{ std::to_underlying(command) };
+      static constexpr std::size_t reverse_bit =
+          11;  // from https://iportal2.schneider-electric.com/Contents/docs/SQD-ATV320U11N4C_USER%20GUIDE.PDF
+      command_word.set(reverse_bit, true);
+      out->command_word = static_cast<uint16_t>(command_word.to_ulong());
+    } else {
+      out->command_word = static_cast<uint16_t>(command);
+    }
+
+    out->frequency = running_ ? reference_frequency_.value.numerical_value_ : 0;
```
omarhogni, line 628: This looks expensive. We have a control_word type bit field that is returned from the cia_402 header. Could we not destroy that type, modify the reserved bit on the bit field and then utilize a std::bit_cast before copying the 16-bit integer to the wire.

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320.hpp
```diff
@@ -396,9 +397,67 @@ using acceleration_ramp_time_ACC = setting<ecx::index_t{ 0x203c, 0x02 }, "ACC",
...
+constexpr auto percentage_to_deci_freq(mp_units::quantity<mp_units::percent, double> percentage,
+                                       [[maybe_unused]] low_speed_LSP min_freq,
+                                       [[maybe_unused]] high_speed_HSP max_freq) noexcept -> speed {
+  if (percentage == 0 * mp_units::percent) {
+    return { .value = 0 * dHz, .reverse = false };
+  }
+  bool reverse = percentage.numerical_value_ < 0;
+  auto mapped{ ec::util::map(mp_units::abs(percentage).numerical_value_, (1.0 * mp_units::percent).numerical_value_,
+                             (100.0 * mp_units::percent).numerical_value_, min_freq.value.numerical_value_,
+                             max_freq.value.numerical_value_) };
```
omarhogni, line 416: Do you need to throw away the types?

omarhogni (changes requested):

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320.hpp
jbbjarnason: since we are mapping cross dimensions, yes. The correct way would be to ask mp to provide map function in his math file

jbbjarnason: Or write explicit mapping function for mp

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320.hpp
jbbjarnason: refactored

exes/ethercat/inc/public/tfc/cia/402.hpp
```diff
@@ -28,6 +28,25 @@ struct control_word {
...
+                         .operating_mode_specific_2 = static_cast<bool>(word & 0b1000000),
+                         .fault_reset = static_cast<bool>(word & 0b10000000),
+                         .halt = static_cast<bool>(word & 0b100000000),
+                         .operating_mode_specific_3 = static_cast<bool>(word & 0b1000000000),
+                         .reserved_0 = static_cast<bool>(word & 0b10000000000),
+                         .reserved_1 = static_cast<bool>(word & 0b100000000000),
+                         .reserved_2 = static_cast<bool>(word & 0b1000000000000),
+                         .reserved_3 = static_cast<bool>(word & 0b10000000000000),
+                         .reserved_4 = static_cast<bool>(word & 0b100000000000000),
+                         .reserved_5 = static_cast<bool>(word & 0b1000000000000000) };
```
omarhogni, line 47:
```suggestion
    return std::bit_cast<control_word>(word);
```

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320.hpp
omarhogni: I think that our static_cast to double might be messing it up but the from type should cancel out with the division in map.

omarhogni (changes requested):

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320.hpp
jbbjarnason: what was the reason for static_cast to double, do you remember?

omarhogni: mp-units uses the underlying types for calculations. You can get values far from what you expect. We can experiment with changing it back though or maybe cast to double and then cast back.

exes/ethercat/inc/public/tfc/cia/402.hpp
jbbjarnason: NICE !

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320.hpp
jbbjarnason: I will tinker with removing the casting and see if it provides same results test wise

exes/ethercat/inc/public/tfc/cia/402.hpp
jbbjarnason: But this is not implemented with libc++ as constexpr, as can be seen in pipelines.
Artur has implementation that I would like to integrate to https://github.com/arturbac/small_vectors/blob/master/include/utils/utility_cxx20.h#L198
I would like to officially finish integrating this repo(small_vectors) with vcpkg as dependency so we can utilize this and updated version of fixed_string, @omarhogni any counters?

omarhogni: This does not make sense to me https://reviews.llvm.org/rGb1fb3d75c953fa2e02ebddb6ebbf100f99786f0c bit_cast is constexpr under libc++ and was added in c++20 as constexpr. Are you sure there is not another reason?

jbbjarnason: they have tests for this error being thrown https://github.com/llvm/llvm-project/blob/2e6f722b88e42bba72af5bafd8cc1e37e460d978/clang/test/SemaCXX/constexpr-builtin-bit-cast.cpp#L143

I can disable the tests on clang

omarhogni: > But this is not implemented with libc++ as constexpr, as can be seen in pipelines. Artur has implementation that I would like to integrate to https://github.com/arturbac/small_vectors/blob/master/include/utils/utility_cxx20.h#L198 I would like to officially finish integrating this repo(small_vectors) with vcpkg as dependency so we can utilize this and updated version of fixed_string, @omarhogni any counters?

I don't think this will be constexpr under clang. We are stuck for now. This is more complex than using another implementation see notes under https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0476r2.html  especially
```
Furthermore, it is currently impossible to implement a constexpr bit-cast function, as memcpy itself isn’t 
constexpr. Marking the proposed function as constexpr doesn’t require or prevent memcpy from becoming 
constexpr, but requires compiler support.
```

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320.hpp
jbbjarnason: now explicitly for double (implicit casting), and mp units truncating to double

exes/ethercat/inc/public/tfc/cia/402.hpp
jbbjarnason: disabled on clang

PR #250, opened 2023-11-08
ethercat path fixed to correct binary path

magni-mar:
ethercat binary path is slightly wrong in .service file for ethercat

### DIFF
--- a/exes/ethercat/CMakeLists.txt
+++ b/exes/ethercat/CMakeLists.txt
@@ -42,7 +42,7 @@ if (BUILD_TESTING)
   add_subdirectory(tests)
 endif ()
 
-configure_file("${CMAKE_SOURCE_DIR}/exes/ethercat/systemd/ethercat@.service" "systemd/ethercat@.service")
+configure_file("${CMAKE_CURRENT_LIST_DIR}/systemd/ethercat@.service" "${CMAKE_BINARY_DIR}/systemd/ethercat@.service")
 
 include(GNUInstallDirs)
 install(
--- a/exes/ethercat/CMakeLists.txt
+++ b/exes/ethercat/CMakeLists.txt
@@ -42,15 +42,17 @@ if (BUILD_TESTING)
   add_subdirectory(tests)
 endif ()
 
+include(GNUInstallDirs)
+
 configure_file("${CMAKE_CURRENT_LIST_DIR}/systemd/ethercat@.service" "${CMAKE_BINARY_DIR}/systemd/ethercat@.service")
 
-include(GNUInstallDirs)
 install(
   FILES
-    systemd/ethercat@.service
+    "${CMAKE_BINARY_DIR}/systemd/ethercat@.service"
   DESTINATION
     /usr/lib/systemd/system/
-  CONFIGURATIONS Release
+  CONFIGURATIONS
+    Release
 )
 
 install(
--- a/exes/ethercat/CMakeLists.txt
+++ b/exes/ethercat/CMakeLists.txt
@@ -42,6 +42,8 @@ if (BUILD_TESTING)
   add_subdirectory(tests)
 endif ()
 
+configure_file("${CMAKE_SOURCE_DIR}/exes/ethercat/systemd/ethercat@.service" "systemd/ethercat@.service")
+
 include(GNUInstallDirs)
 install(
   FILES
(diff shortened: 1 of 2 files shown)


exes/ethercat/systemd/ethercat@.service
```diff
@@ -4,7 +4,7 @@ Requires=dbus.socket
 After=dbus.socket
 
 [Service]
-ExecStart=/usr/bin/ethercat --iface %i
+ExecStart=/usr/local/bin/ethercat --iface %i
```
omarhogni, line 7: Can you template this on CMAKE_INSTALL_BINDIR so it does not break again?

omarhogni (changes requested):

exes/ethercat/systemd/ethercat@.service
magni-mar: Will do, the variable `CMAKE_INSTALL_PREFIX` is already used in another place in the repo so using that variable.

jbbjarnason (changes requested):
I think you are missing a crucial step if you are going to use variable from cmake. Have you manually tested this and did it work?

magni-mar:
> I think you are missing a crucial step if you are going to use variable from cmake. Have you manually tested this and did it work?

I have manually tested this and it did work, what crucial step do you think I am missing?

jbbjarnason:
> > I think you are missing a crucial step if you are going to use variable from cmake. Have you manually tested this and did it work?
> 
> I have manually tested this and it did work, what crucial step do you think I am missing?

Here is a hint,

image


It would be nice to know how you tested it, because I am pretty confident that CMAKE_INSTALL_PREFIX in systemd file will not work.

magni-mar:
> > > I think you are missing a crucial step if you are going to use variable from cmake. Have you manually tested this and did it work?
> > 
> > 
> > I have manually tested this and it did work, what crucial step do you think I am missing?
> 
> Here is a hint,
> 
> image
> 
> It would be nice to know how you tested it, because I am pretty confident that CMAKE_INSTALL_PREFIX in systemd file will not work.

I understand what you mean, after looking into it I added `configure_file` which is similar to what is done in `tfc_systemd`, a step which I forgot to include.

exes/ethercat/CMakeLists.txt
```diff
@@ -42,6 +42,8 @@ if (BUILD_TESTING)
   add_subdirectory(tests)
 endif ()
 
+configure_file("${CMAKE_SOURCE_DIR}/exes/ethercat/systemd/ethercat@.service" "systemd/ethercat@.service")
```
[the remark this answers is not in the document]
magni-mar: It doesn't override the local file but places it in another directory in the build file, to make this more clear I will add `${CMAKE_BINARY_DIR}/systemd/ethercat@.service` so it will be placed with other systemd files in the build directory.

Will add  `${CMAKE_CURRENT_LIST_DIR}/systemd/ethercat@.service` to remove the absolute path dependancy.

exes/ethercat/CMakeLists.txt
```diff
@@ -42,6 +42,8 @@ if (BUILD_TESTING)
   add_subdirectory(tests)
 endif ()
 
+configure_file("${CMAKE_CURRENT_LIST_DIR}/systemd/ethercat@.service" "${CMAKE_BINARY_DIR}/systemd/ethercat@.service")
+
```
jbbjarnason, line 46: Does the install in line 50 need to point to binary for?

magni-mar: https://cmake.org/cmake/help/latest/variable/CMAKE_BINARY_DIR.html#variable:CMAKE_BINARY_DIR

this variable points to the top level in the build directory, do you have any other suggestion?

omarhogni (approved):

exes/ethercat/CMakeLists.txt
jbbjarnason: I cannot place a comment on line 50 on github, the reason I mentioned it directly. The install is broken because of this configure_file, and needs to be fixed there. Since you now place it in binary dir you need to install from binary dir.
 If you test the installer, you will see the service file is still the same:
image

magni-mar: Changed the path in the installer, this should work now.

jbbjarnason (approved):

PR #259, opened 2023-11-13
Add checkbox to subscribe and unsubscribe from slots andn signals

### LINKED ISSUE #258: Io debugging UI heavy load on backend
image

Maybe we should cherry pick each signal/slot to subscribe to. Checkmark each to watch

SindriTh:
fixes #258 

Added checkboxes to DebugUI to subscribe and unsubscribe from changes for that particular item.

@jbbjarnason Could you check if this reduces load?

### DIFF
--- a/cockpit/IPC/src/Components/IOList/ListItem.tsx
+++ b/cockpit/IPC/src/Components/IOList/ListItem.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/no-unstable-nested-components */
 /* eslint-disable no-param-reassign */
-import React, { ReactElement } from 'react';
+import React, { ChangeEvent, ReactElement } from 'react';
 import {
   ClipboardCopy,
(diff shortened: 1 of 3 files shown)


omarhogni:
Should we not hide the values and indicators when values are unwatched. We also need some method to select allot after having filterered. f.e. I filter EL2004 and want to monitor the entire result range can be kind of time consuming. Can we shift-select?

SindriTh:
> Should we not hide the values and indicators when values are unwatched. We also need some method to select allot after having filterered. f.e. I filter EL2004 and want to monitor the entire result range can be kind of time consuming. Can we shift-select?

Selecting using keyboard shortcut is not currently available.
I was wondering about hiding them as well, should we do that as a filter? that sounds like the most reasonable option.
We would need a way to add them to the watchlist

jbbjarnason:
disable fields and tinkering if untoggled

PR #260, opened 2023-11-13
Fixed null handling on dbus

SindriTh:
Fix handling of nulls on DBUS

Error came up when dbus value was a non-nested straight null


omarhogni (approved):
LGTM

PR #280, opened 2023-11-20
add or and either guard cases to sml logger

magni-mar:


jbbjarnason (approved):
lgtm I will do some minor cleanup later

PR #293, opened 2023-11-27
static used in constexpr context

omarhogni:


jbbjarnason:
@omarhogni, what compiler is erroring, this should be supported by our toolchains according to 

image

jbbjarnason (approved):
If this is a blocker somewhere please do merge, but I find this to be weird

omarhogni:
clang causes issues.

PR #295, opened 2023-11-27
Fix segfault in mqtt bridge

magni-mar:
In order to reduce CPU load #302 is necessary


libs/ipc/inc/public/tfc/ipc/details/impl.hpp
```diff
@@ -299,11 +305,23 @@ class slot_callback : public std::enable_shared_from_this<slot_callback<type_des
   using value_t = type_desc::value_t;
   static auto constexpr direction_v = slot<type_desc>::direction_v;
 
+  auto type() const -> const type_desc& {
+    static type_desc type;
+    return type;
+  }
```
jbbjarnason, line 311: please don`t

libs/ipc/inc/public/tfc/ipc/details/impl.hpp
jbbjarnason, line 1: why is it needed for underlying impl of ipc to store `description`

jbbjarnason (approved):

jbbjarnason (changes requested):
ipc changes needs to be discussed

exes/mqtt-bridge/inc/config/broker.hpp
```diff
@@ -3,37 +3,15 @@
...
-  static constexpr auto value = glz::enumerate("yes", yes, "no", no);
-};
-
 namespace tfc::mqtt::config {
 
 struct broker {
   std::string address{};
-  std::variant<port_e, uint16_t> port{};
-  structs::ssl_active_e ssl_active{};
+  uint16_t port{};
```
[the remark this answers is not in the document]
magni-mar: variant was not loading correctly from the config file, this is possibly related to the update of glaze

libs/ipc/inc/public/tfc/ipc/details/impl.hpp
magni-mar: I need some way of getting the type of the slot, alternative suggestion are welcome

exes/mqtt-bridge/inc/spark_plug_interface.hpp
```diff
@@ -99,7 +99,7 @@ class spark_plug_interface {
   asio::io_context& io_ctx_;
   config_t config_{ io_ctx_, "spark_plug_b_config" };
   std::unique_ptr<mqtt_client_t> mqtt_client_;
-  std::vector<structs::signal_data> nbirth_;
+  std::vector<tfc::ipc::details::any_slot_cb> nbirth_;
```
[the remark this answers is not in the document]
magni-mar: changed the name to `variables_` and changed the data type as well

exes/mqtt-bridge/inc/spark_plug_interface.hpp
```diff
@@ -62,14 +60,16 @@ class spark_plug_interface {
...
-  auto set_current_values(std::vector<structs::signal_data> const& metrics) -> void;
+  auto set_current_values(std::vector<tfc::ipc::details::any_slot_cb> const& metrics) -> void;
 
   auto send_current_values() -> void;
 
-  auto update_value(structs::signal_data& signal) -> asio::awaitable<void>;
-
   auto make_payload() -> Payload;
 
+  auto update_value(tfc::ipc::details::any_slot_cb& signal_data) -> void;
```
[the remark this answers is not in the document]
magni-mar: updated this parameter name to `variable` along with the data type

exes/mqtt-bridge/inc/spark_plug_interface.hpp
```diff
@@ -62,14 +60,16 @@ class spark_plug_interface {
 
   static auto timestamp_milliseconds() -> std::chrono::milliseconds;
 
-  auto set_current_values(std::vector<structs::signal_data> const& metrics) -> void;
+  auto set_current_values(std::vector<tfc::ipc::details::any_slot_cb> const& metrics) -> void;
```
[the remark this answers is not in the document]
magni-mar: updated the parameter name along with data type

exes/mqtt-bridge/inc/tfc_to_external.hpp
```diff
@@ -1,80 +1,95 @@
...
-#include <config/spark_plug_b.hpp>
-#include <config/spark_plug_b_mock.hpp>
-#include <spark_plug_interface.hpp>
-#include <structs.hpp>
-#include <tfc/confman.hpp>
 #include <tfc/ipc.hpp>
-#include <tfc/ipc/details/dbus_client_iface_mock.hpp>
 #include <tfc/logger.hpp>
 
+#include <spark_plug_interface.hpp>
```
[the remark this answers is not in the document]
magni-mar: do you have an alternative suggestion

exes/mqtt-bridge/src/spark_plug_interface.cpp
```diff
@@ -40,6 +40,79 @@ auto spark_plug_interface<config_t, mqtt_client_t>::get_bd_seq() -> int64_t {
...
+  }
+
+  metric->set_name(variable);
+  metric->set_timestamp(timestamp_milliseconds().count());
+  metric->set_datatype(type_enum_convert(type));
+
+  std::visit(
+      [this, &metric]<typename recv_t>(recv_t&& receiver) {
+        using receiver_t = std::remove_cvref_t<decltype(receiver)>;
+        if constexpr (std::same_as<receiver_t, ipc::details::bool_slot_cb_ptr>) {
```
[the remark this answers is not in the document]
magni-mar: I have removed the dependancy that spark plug interface depends on ipc layer.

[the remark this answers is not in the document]
magni-mar: general ipc to external needs some api to update the variable in the external interface

libs/ipc/inc/public/tfc/ipc/details/impl.hpp
magni-mar: this is also outdated, variable information is kept in a specified struct

libs/ipc/inc/public/tfc/ipc/details/impl.hpp
magni-mar: I have removed that dependancy

exes/mqtt-bridge/inc/tfc_to_external.hpp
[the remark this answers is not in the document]
magni-mar: it's possible to depend on a generic interface in which spark plug b or possibly other communication protocols could depend upon, tfc to external would then depend upon that interface but in my opinion this is overengineering since we are only using spark plug b, it would be a good idea as soon as another external communication protocol is introduced

exes/mqtt-bridge/inc/structs.hpp
```diff
@@ -1,18 +1,15 @@
...
-
-enum struct ssl_active_e { yes, no };
-
-struct signal_data {
-  tfc::ipc_ruler::signal information;
-  tfc::ipc::details::any_slot receiver;
-  std::optional<std::any> current_value;
+struct spark_plug_b_variable {
+  std::string name;
+  org::eclipse::tahu::protobuf::DataType datatype;
```
jbbjarnason, line 12: is it needed to store the datatype here? can it not be deduced from the value itself?

exes/mqtt-bridge/inc/structs.hpp
```diff
@@ -1,18 +1,15 @@
...
+  std::optional<std::any> value;
```
jbbjarnason, line 13: I would suggest storing value as variant.

jbbjarnason (commented):
I would suggest to revert the configuration change, do not apply any more big changes. The topic of this PR was to fix the segfault but not refactor the whole thing.

exes/mqtt-bridge/inc/structs.hpp
magni-mar: if the optional is empty then it is impossible to know what the datatype is. I also don´t see a problem with having the struct maintain all needed information about each variable

exes/mqtt-bridge/inc/structs.hpp
magni-mar: Then it is impossible to know weather the value is of json type or string type, as they are both `std::string`

magni-mar:
> I would suggest to revert the configuration change, do not apply any more big changes. The topic of this PR was to fix the segfault but not refactor the whole thing.

the old configuration produces the following error:

```bash
*** warning [00:01:35 +00:00] (thread 92356) def.file_storage./etc/tfc/mqtt_bridge/def/client.json *** 		 Error: ""no_matching_variant_type"" reading from file: "/etc/tfc/mqtt_bridge/def/client.json" 
```

magni-mar:
If the old configuration is to be used then #324 needs to be resolved first.

jbbjarnason:
> If the old configuration is to be used then #324 needs to be resolved first.

It works with a number,

image

magni-mar:
> > > If the old configuration is to be used then #324 needs to be resolved first.
> > 
> > 
> > It works with a number,
> > image
> 
> There seems to be a fix on glaze so either wait for an update on glaze or change to number and merge. In my opinion the configuration should be changed to number and when a new release is dropped the config can be changed.

@jbbjarnason @omarhogni  config will stay as is and wait for glaze update, until then the field in the config just needs to be a number. a few files were checked out from previous version so this PR is less of a change

jbbjarnason (approved):

PR #298, opened 2023-11-29
Debugui build flag

SindriTh:
Need help changing the CMakeLists file to allow for a build flag for DebuggingUI.

@jbbjarnason


jbbjarnason:
@SindriTh would it not be sufficient to default off when built with CMAKE, but default on if you run npm run devbuild?

SindriTh:
> @SindriTh would it not be sufficient to default off when built with CMAKE, but default on if you run npm run devbuild?
> 
> As in not optional in cmake

We could also do that yes

jbbjarnason (approved):

PR #305, opened 2023-12-01
Added const to store every time, and added ReadOnly to Units dropdown

SindriTh:


SindriTh:
Added validation checking, as it had a tendency to fail with the new setup.
Feel free to break it @jbbjarnason.

I have yet to figure out a nice way to handle arrays, this might be the next project on my todo. That would likely fix the visual glitches when removing an item, as well as allowing for hard limits for item counts.

That would of course be a separate PR.

jbbjarnason (approved):

PR #307, opened 2023-12-01
Added workarounds to be able to include the project from another project

omarhogni:


jbbjarnason (approved):
I don't like this but lets keep it rollin

magni-mar (approved):
lgtm

PR #315, opened 2023-12-04
vcpkg install error

magni-mar:


omarhogni (commented):
I would like to close this in favour of #316

PR #317, opened 2023-12-05
Motor interface continuation

omarhogni:


CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ include(tfc_config)
 include(tfc_warnings)
 include(tfc_docs)
 include(tfc_cpack)
+include(tfc_systemd)
```
jbbjarnason, line 24: wasn't this in previous PR?

libs/motor/inc/public/tfc/motors/mock_motor.hpp
```diff
@@ -0,0 +1,55 @@
...
+ * @brief mock motor class
+ * @details This struct is only used for testing purposes.
+ * none of the methods should be called in production.
+ * This struct is also simple and does not rely on time by
+ * default.
+ */
+
+template <typename clock_t>
+struct mock_motor {
+  explicit virtual_motor(boost::asio::io_context&) {}
```
jbbjarnason, line 29: discrepancy here, does it compile?

libs/motor/inc/public/tfc/motors/virtual_motor.hpp
```diff
@@ -0,0 +1,194 @@
...
+    if (!config_.nominal) {
+      return motor_error(errors::err_enum::motor_missing_speed_reference);
+    }
+    [[maybe_unused]] auto frequency = tfc::motor::impl::nominal_at_50Hz_to_frequency(config_.nominal.value(), vel);
+    logger_.info("convey running at {}", frequency);
+    running_ = true;
+    return {};
+  }
+
+  void convey(QuantityOf<mp_units::isq::velocity> auto vel,
```
jbbjarnason, line 77: What should occur if convey is called twice in the same event cycle?

libs/motor/inc/public/tfc/motors/virtual_motor.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+  ~virtual_motor() {}
+
+  auto convey() -> std::error_code {
+    logger_.info("convey!");
+    running_ = true;
+    return {};
+  }
+
+  auto convey(QuantityOf<mp_units::isq::velocity> auto vel) -> std::error_code {
```
jbbjarnason, line 66: maybe look into making one private `convey` function which could cover all public instances, less duplication. could use non type template parameters to activate or deactivate code

CMakeLists.txt
omarhogni: Rebased.

libs/motor/inc/public/tfc/motors/mock_motor.hpp
omarhogni: This idea is not included anywhere i believe. Lets revisit when there are some tests written on top of this and in usage.

libs/motor/inc/public/tfc/motors/virtual_motor.hpp
omarhogni: If you mean someone calls convey when the conveyor is already running?
We had discussed that the new parameters would overtake the previous invocation.

f.e.
```cpp
convey(5m)
// Conveyor will move 10meters from the point it had reached in completing the 5m transition
convey(10m) 
```

omarhogni: exactly. However all of this needs more maturity and testing. Especially for these edge cases.

libs/motor/inc/public/tfc/motor/atv320motor.hpp
```diff
@@ -0,0 +1,219 @@
+#pragma once
+#include <mp-units/chrono.h>
+
+#include <iostream>  //TODO: Remove
```
jbbjarnason, line 4: remove?

libs/stx/inc/public/tfc/utils/units_glaze_meta.hpp
```diff
@@ -61,6 +95,7 @@ struct to_json_schema<mp_units::quantity<ref_t, rep_t>> {
     if (!data.has_value()) {
       data = tfc::json::schema_meta{};
     }
+    assert(unit_ascii[0] != '\0');
```
jbbjarnason, line 98:
```suggestion
```

jbbjarnason (approved):

PR #321, opened 2023-12-05
read only propagated downwards

jbbjarnason:


SindriTh (approved):

omarhogni (approved):
Is this sufficient? 

you cannot have f.e. read only std::maps using this

jbbjarnason:
> Is this sufficient?
> 
> you cannot have f.e. read only std::maps using this

you are correct, will do those additions (also applies to vectors/array)

libs/stx/inc/public/tfc/utils/json_schema.hpp
```diff
@@ -437,6 +439,9 @@ struct to_json_schema<T> {
       using V = std::decay_t<std::variant_alternative_t<I, T>>;
       auto& schema_val = (*s.oneOf)[I.value];
       schema_val.attributes.title = glz::name_v<V, true>;  // please MAKE sure you declare name within variants
+      if (s.attributes.read_only.has_value()) {
+        schema_val.attributes.read_only = s.attributes.read_only;
+      }
```
omarhogni, line 444:
```suggestion
schema_val.attributes.read_only = s.attributes.read_only;
```
```suggestion
      schema_val.attributes.read_only = s.attributes.read_only;
```

omarhogni (approved):
It would be good to see some trivial tests for this. could encode a json schema and parse it back as normal json. Then the fields could be checked.

jbbjarnason:
let's close for now this is a bit trickier than I thought

PR #327, opened 2023-12-06
Glaze bug with std::variant

magni-mar:


omarhogni:
@magni-mar can we close this?

magni-mar:
> @magni-mar can we close this?

yes

PR #335, opened 2023-12-08
vcpkg update

jbbjarnason:


jbbjarnason:
Followup issue https://github.com/stephenberry/glaze/issues/604

omarhogni:
The removal of `std::chrono::steady_clock` from the glaze_meta test is caused by glazes ability to reflect the type and come up with it's own encoding.

magni-mar (approved):
lgtm

PR #337, opened 2023-12-08
Positioning for conveyors

jbbjarnason:


libs/motor/inc/public/tfc/motors/positioner.hpp
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  struct storage {
+    bool first_tacho_state{};
+    bool second_tacho_state{};
+    time_point_t time_point{};
+  };
+  circular_buffer<storage, circular_buffer_len> buffer_{};
+  std::int64_t position_{};
+  std::optional<std::function<void(std::int64_t)>> position_update_callback_;
```
omarhogni, line 218:
```suggestion
  std::function<void(std::int64_t)> position_update_callback_ = [](std::int64_t) {};
```

libs/motor/inc/public/tfc/motors/positioner.hpp
```diff
@@ -0,0 +1,335 @@
...
+
+    auto now{ time_point_t::clock::now() };
+    buffer_.emplace(buffer_.front().first_tacho_state, second_new_val, now);
+  }
+
+  auto increment() noexcept -> void {
+    position_ += 1;
+    if (position_update_callback_) {
+      (*position_update_callback_)(position_);
+    }
```
omarhogni, line 201:
```suggestion
      position_update_callback_(position_);
```

libs/motor/inc/public/tfc/motors/positioner.hpp
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  auto decrement() noexcept -> void {
+    position_ -= 1;
+    if (position_update_callback_) {
+      (*position_update_callback_)(position_);
+    }
```
omarhogni, line 208:
```suggestion
      position_update_callback_(position_);
```

libs/motor/inc/public/tfc/motors/positioner.hpp
```diff
@@ -0,0 +1,335 @@
...
+      if (!buffer_.front().second_tacho_state) {
+        if (second_new_val) {
+          increment();
+        }
+      } else {
+        if (!second_new_val) {
+          decrement();
+        }
+      }
+    }
```
omarhogni, line 191: This is hard to read

omarhogni: Can we not merge the data and utilize a similar function as the update function in the comments. It is also allot faster.

omarhogni (changes requested):

libs/motor/inc/public/tfc/motors/positioner.hpp
```diff
@@ -0,0 +1,335 @@
...
+                                  position -= 2; break;
+                  }
+                  state = (s >> 2);
+          }
+  */
+
+  void first_tacho_update(bool first_new_val) noexcept {
+    if (buffer_.front().first_tacho_state) {
+      if (buffer_.front().second_tacho_state) {
+        if (!first_new_val) {
```
jbbjarnason, line 146: I would not consider falling edge reliable source of information, if you want to keep it a duty cycle parameter needs to be added.

jbbjarnason: Duty cycle can be skewed, therefore, each increment can mean different length.

magni-mar: the length of each increment is configurable throught `displacement_per_pulse_` variable

jbbjarnason: time of high and low state is not strictly same, but let's not make this stop the PR. The error would be at maximum less than displacement per pulse.
When we need high resolution placement we could implement timers between edges and increase the resolution with that way, then the duty cycle would be determined on the fly.

jbbjarnason (commented):
I would suggest merging this PR, I will make a followup PR going back to completion token. It will simplify the things that build on top of this, as they would be able to support completion token in simple manner.

PR #344, opened 2023-12-18
Add filter declaration for mass_t

magni-mar:
Constructor failed for mass slot, so added a small test to check the constructor so that this is prevented in the future. Packet deserialization has changed according to mass error, this needs to be discussed and reviewed thoroughly.


examples/mass_example/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+  tfc::ipc::slot<tfc::ipc::details::type_mass, tfc::ipc_ruler::ipc_manager_client_mock&> weight_slot{
+    ctx, ipc_client_, "weight", [](const tfc::ipc::details::mass_t&) {}
+  };
+
+  ipc_client_.connect(ipc_client_.slots_[0].name, ipc_client_.signals_[0].name, [](std::error_code const&) {});
+
+  weight_signal.send(100 * mp_units::si::gram);
+
+  ctx.run_for(std::chrono::milliseconds(10));
```
jbbjarnason, line 34: looks like ping pong test, It would be awesome to convert this to a ipc test using parameterized test.

something like:
```cpp
template <typename type_decl>
struct data_t {
  using type_description = type_decl;
  type_decl::value_t value{};
};
"ping pong"_test = [](auto& data) {
  using type_description = std::remove_ref_t<decltype(data)>::type_description;
  // make signal and slot connect and send and receive and expect upon its value
} | std::tuple{ data_t<tfc::ipc::details::type_mass>{.value = 100 * mp_units::si::gram }  }; // add all types here with its according value
```

libs/ipc/inc/public/tfc/ipc/details/filter.hpp
```diff
@@ -247,6 +248,11 @@ struct any_filter_decl<std::string> {
   using value_t = std::string;
   using type = std::variant<filter<filter_e::filter_out, value_t>>;
 };
+template <>
+struct any_filter_decl<ipc::details::mass_t> {
+  using value_t = ipc::details::mass_t;
+  using type = std::variant<filter<filter_e::filter_out, value_t>>;
```
jbbjarnason, line 254: This is how I would like this. 

But I am not sure whether adding this filter would actually work, can you test it? If doesn't work open a followup issue.

libs/ipc/inc/public/tfc/ipc/packet.hpp
```diff
@@ -32,7 +32,9 @@ struct header_t {
     std::copy_n(reinterpret_cast<std::byte*>(&header.type), sizeof(type), std::back_inserter(buffer));
     std::copy_n(reinterpret_cast<std::byte*>(&header.value_size), sizeof(value_size), std::back_inserter(buffer));
   }
-  static auto deserialize(header_t& result, auto&& buffer_iter) -> std::error_code {
+
+  template <typename error_t = std::error_code>
+  static auto deserialize(header_t& result, auto&& buffer_iter) -> error_t {
```
jbbjarnason, line 37: I think this is unnecessary, and does not work for anything other than std error_code

libs/ipc/inc/public/tfc/ipc/packet.hpp
```diff
@@ -118,7 +120,8 @@ struct packet {
     return {};
   }
 
-  static constexpr auto deserialize(std::ranges::view auto&& buffer) -> std::expected<value_t, std::error_code> {
+  template <typename error_t = std::error_code>
+  static constexpr auto deserialize(std::ranges::view auto&& buffer) -> std::expected<value_t, error_t> {
```
jbbjarnason, line 124: This is also unnecessary.
This can be a bit mind boggling, in case of mass the value_t is std::expected so the return type is std::expected<std::expected<mass_t, mass_error_e>, std::error_code>

libs/ipc/inc/public/tfc/ipc/packet.hpp
```diff
@@ -148,9 +151,13 @@ struct packet {
...
-        typename value_t::error_type substitute{};
-        std::copy_n(buffer_iter, result.header.value_size - 1, reinterpret_cast<std::byte*>(&substitute));
-        result.value = std::unexpected{ substitute };
+        if constexpr (std::is_same_v<typename value_t::error_type, mass_error_e>) {
+          result.value = std::unexpected{ static_cast<mass_error_e>(*buffer_iter) };
+        } else {
+          typename value_t::error_type substitute{};
+          std::copy_n(buffer_iter, result.header.value_size - 1, reinterpret_cast<std::byte*>(&substitute));
+          result.value = std::unexpected{ substitute };
+        }
```
jbbjarnason, line 160: I don't like this being explicitly done for this error enum, how is this broken like it is now?

jbbjarnason (changes requested):

libs/ipc/inc/public/tfc/ipc/packet.hpp
magni-mar: https://github.com/Skaginn3x/framework/actions/runs/7248918924/job/19745806361

this is the error that happens, 

```bash
 forming offset 1 is out of the bounds [0, 1] of object 'substitute' with type 'std::expected<mp_units::quantity<mp_units::si::milli_<mp_units::si::gram()>(), long int>, tfc::ipc::details::mass_error_e>::error_type' {aka 'tfc::ipc::details::mass_error_e'}
```

jbbjarnason: Is this only an error for gcc release mode?
Is this a warning or an actual error?

examples/mass_example/main.cpp
magni-mar: I added this test but that doesn't seem to test the failing piece of code, still nice to have.

libs/ipc/inc/public/tfc/ipc/packet.hpp
magni-mar: This is only an error for gcc-release, it is an error, not warning.

libs/ipc/inc/public/tfc/ipc/details/filter.hpp
magni-mar: Added a test for the filter, it works.

magni-mar:
I added an offset filter along with the out filter, this also needs to be discussed if it is applicable or desireable for this type.

jbbjarnason:
image

This just works, nice.

I would prefer not having offset, you would be tampering with value from the weigher which is not something I think we would like to do so easily.

libs/ipc/inc/public/tfc/ipc/details/filter.hpp
```diff
@@ -247,6 +255,11 @@ struct any_filter_decl<std::string> {
   using value_t = std::string;
   using type = std::variant<filter<filter_e::filter_out, value_t>>;
 };
+template <>
+struct any_filter_decl<ipc::details::mass_t> {
+  using value_t = ipc::details::mass_t;
+  using type = std::variant<filter<filter_e::filter_out, value_t>, filter<filter_e::offset, value_t>>;
```
jbbjarnason, line 261: like said in comment I do not like the offset being an option here

libs/ipc/testing/tests/ipc_test.cpp
```diff
@@ -142,5 +150,38 @@ auto main(int argc, char** argv) -> int {
...
+        expect(value_received);
+      } |
+      std::tuple{ data_t<tfc::ipc::details::type_mass>{ .value = 100 * mp_units::si::gram },
+                  data_t<tfc::ipc::details::type_bool>{ .value = true },
+                  data_t<tfc::ipc::details::type_double>{ .value = 3.14 },
+                  data_t<tfc::ipc::details::type_int>{ .value = 3 },
+                  data_t<tfc::ipc::details::type_json>{ .value =
+                                                            R"({"i":287,"d":3.14,"hello":"Hello World","arr":[1,2,3])" },
+                  data_t<tfc::ipc::details::type_string>{ .value = "hello world from another world" },
+                  data_t<tfc::ipc::details::type_uint>{ .value = std::numeric_limits<std::uint64_t>::max() } };
```
jbbjarnason, line 184: Nice tests,
Can you add one error case for mass type?

jbbjarnason (commented):
I committed changes to packet which compile on gcc release

libs/ipc/inc/public/tfc/ipc/details/filter.hpp
magni-mar: I'll remove it, don't have a strong opinion on this.

libs/ipc/testing/tests/ipc_test.cpp
magni-mar: Added an error case.

libs/ipc/testing/tests/ipc_test.cpp
```diff
@@ -142,5 +150,40 @@ auto main(int argc, char** argv) -> int {
...
+          }
+        };
+        ipc_client.connect(ipc_client.slots_[0].name, ipc_client.signals_[0].name, [](std::error_code const&) {});
+        ctx.run_for(std::chrono::milliseconds(5));
+        sender.send(data.value);
+        ctx.run_for(std::chrono::milliseconds(5));
+        expect(value_received);
+      } |
+      std::tuple{ data_t<type_mass>{ .value = 100 * mp_units::si::gram },
+                  data_t<type_mass>{ .value = std::unexpected(mass_error_e::cell_fault) },
```
jbbjarnason, line 180: nice!

jbbjarnason (approved):
Looks good, squash and merge it

PR #345, opened 2023-12-18
MQTT signals whitelist

magni-mar:


exes/mqtt-bridge/inc/tfc_to_external.hpp
```diff
@@ -85,6 +98,10 @@ class tfc_to_external {
     spb_variables_.reserve(signals.size());
 
     for (const auto& signal : signals) {
+      if (!is_publish_signal(signal.name)) {
```
jbbjarnason, line 101: if the list is empty should we publish everything?

magni-mar: If we do that I suspect that the feature would be not used, and CPU load would always be higher than strictly necessary. If nothing is published for an empty list then the user is forced to insert the necessary signals.

jbbjarnason:
This will be quite HEAVY to configure, and I would like to somehow make it easier to configure.

My suggestion is writing custom schema and populate to the UI the available options. Let's explain it in code.

FYI did not try to compile this
```cpp
struct signal_name {
  std::string value{};
};

static std::vector<std::string> get_available_signal_names() {
  asio::io_context ctx{};
  ipc_manager_client client{ ctx };
  auto names = client.fetch_signal_names();
  ctx.poll_one();
  return names;
}

template<>
struct tfc::json::to_json_schema<signal_name> {
  template <auto Opts>
  static void op(auto& s, auto& defs) noexcept {
    auto foo = get_available_signal_names();
    s.oneOf = std::vector<schematic>{ { .constant = "signal name 1", .title = "signal name 1" }};
  };
};

```

@SindriTh could we make the one of list filterable in UI if it has more than let's say 10 items in the list?

magni-mar:
> This will be quite HEAVY to configure, and I would like to somehow make it easier to configure.
> 
> My suggestion is writing custom schema and populate to the UI the available options. Let's explain it in code.
> 
> FYI did not try to compile this
> 
> ```c++
> struct signal_name {
>   std::string value{};
> };
> 
> static std::vector<std::string> get_available_signal_names() {
>   asio::io_context ctx{};
>   ipc_manager_client client{ ctx };
>   auto names = client.fetch_signal_names();
>   ctx.poll_one();
>   return names;
> }
> 
> template<>
> struct tfc::json::to_json_schema<signal_name> {
>   template <auto Opts>
>   static void op(auto& s, auto& defs) noexcept {
>     auto foo = get_available_signal_names();
>     s.oneOf = std::vector<schematic>{ { .constant = "signal name 1", .title = "signal name 1" }};
>   };
> };
> ```
> 
> @SindriTh could we make the one of list filterable in UI if it has more than let's say 10 items in the list?

```cpp
struct signal_name {
  std::string value{};
};

static inline std::vector<std::string> get_available_signal_names() {
  boost::asio::io_context ctx{};
  tfc::ipc_ruler::ipc_manager_client client{ ctx };
  std::vector<std::string> names{};
  client.signals([&names](std::vector<tfc::ipc_ruler::signal> const& signals) {
    for (auto const& signal : signals) {
      names.push_back(signal.name);
    }
  });
  ctx.run_for(std::chrono::milliseconds{ 100 });
  return names;
}

template <>
struct tfc::json::detail::to_json_schema<signal_name> {
  template <auto Opts>
  static void op(auto& s, auto& defs) noexcept {
    for (auto const& name : get_available_signal_names()) {
      s.oneOf.push_back(tfc::json::detail::schematic{ .attributes{ tfc::json::schema{ .title = name } } });
    };
  }
};
```

This compiles, can be changed but first iteration. is this added to json_schema file and then configured further in the frontend or do I need to add something more to the mqtt code?

jbbjarnason:
This would be added to the mqtt code. Could be stored where the config relies. 
And 100ms is too much. We cannot block the thread for so long when frontend queries names.

jbbjarnason:
I guess you query this vector somewhere. You could expose it globally. Make the vector as a singleton, and have one method with Io context  and another function to get the last polled vector. This vector would need to be stored in one translation unit, cpp file. Similar to progbase but a bit simpler.

magni-mar:
> I guess you query this vector somewhere. You could expose it globally. Make the vector as a singleton, and have one method with Io context and another function to get the last polled vector. This vector would need to be stored in one translation unit, cpp file. Similar to progbase but a bit simpler.

Storing the struct where the config is, please see newest version. By what mechanism should this be exposed? I am initializing the struct but it doesn't appear on dbus or the front end.

exes/mqtt-bridge/inc/config/publish_signals.hpp
```diff
@@ -0,0 +1,47 @@
...
+    for (auto const& name : get_available_signal_names()) {
+      s.oneOf.push_back(tfc::json::detail::schematic{ .attributes{ tfc::json::schema{ .title = name } } });
+    };
+  }
+};
+
+namespace tfc::mqtt::config {
+
+struct publish_signals {
+  std::vector<std::string> publish_signals{};
```
jbbjarnason, line 39: To populate in frontend

```suggestion
  std::vector<signal_name> publish_signals{};
```
you can add operator to signal_name to implicitly convert the struct to std string if you like.
example:

```cpp
#include <string>

struct signal_name {
  std::string value{};
  constexpr operator std::string_view() const noexcept {
    return value;
  }
};

int main() {
    signal_name name{ .value = "foo" };
    std::string_view foo_view{ name };
    std::string foo_copy{ name };
    return 0;
}
```

exes/mqtt-bridge/inc/tfc_to_external.hpp
```diff
@@ -75,6 +77,17 @@ class tfc_to_external {
         [this](std::vector<ipc_ruler::signal> const& signals) { handle_incoming_signals_from_ipc_client(signals); });
```
jbbjarnason, line 77: line 76 here is doing the same work as the static call in schema, you can expose the output of this call to thread_local vector, but remember to put it into source file similar to this https://github.com/Skaginn3x/framework/blob/main/libs/testing/src/asio_clock.cpp#L6 and expose it via setter and getter

Then you won't need to create the context and block, just use the getter function

magni-mar: I don't understand, the struct still needs to be initialized.

In the clock example it is initialized as such:

```cpp
asio::basic_waitable_timer<tfc::testing::clock> timer{ ctx };
```

making each instance a new instance, and therefore has the same effect as just calling the signals callback on the ipc client.

jbbjarnason: I suggest new header but you can use this one,

in header make free functions
```cpp
namespace global { // please name better
void set_signal_names(std::vector<std::string> const&);
std::vector<std::string> const& get_signal_names();
}
```

source file
```cpp
namespace {
thread_local std::vector<std::string> signal_names{};
}
namespace global { // please name better
void set_signal_names(std::vector<std::string> const& names) {
  signal_names = names;
}
std::vector<std::string> const& get_signal_names() {
  return signal_names;
}
}
```

working example https://godbolt.org/z/73drvvPan

jbbjarnason: Then the following code can be simplified: 
```cpp
template <>
struct tfc::json::detail::to_json_schema<signal_name> {
  template <auto Opts>
  static void op(auto& s, auto& defs) noexcept {
    for (auto const& name : global::get_signal_names()) {
      s.oneOf.push_back(tfc::json::detail::schematic{ .attributes{ tfc::json::schema{ .title = name } } });
    };
  }
};
```

jbbjarnason: and `handle_incoming...` could call `global::set_signal_names(...)`

magni-mar:
Config is capable of being read from but frontend cannot alter the config correctly, it thinks the `signal_name` vector is a vector of `std::string` and changes the values in accordance to that. Not sure where that error lies.

jbbjarnason (approved):

PR #349, opened 2023-12-22
[ethercat] dropdown for network interfaces in ethercat config

magni-mar:


exes/ethercat/inc/tfc/ec/config/ethercat.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+namespace tfc::ec::config {
+struct network_interfaces {
+  network_interface interfaces{ get_network_interfaces()[0] };
+
+  struct glaze {
+    static constexpr auto value{ glz::object("interfaces", &network_interfaces::interfaces, "Network interfaces") };
+    static constexpr std::string_view name{ "network_interfaces" };
+  };
+};
```
jbbjarnason, line 77: I know this is incomplete PR but,

Follow up PRs could add minimum slave count and secondary interface

```suggestion
namespace tfc::ec::config {
struct bus {
  network_interface primary_interface{ get_network_interfaces()[0] };

  struct glaze {
    static constexpr auto value{ glz::object("primary_interface", &bus::primary_interface, "Primary interface") };
    static constexpr std::string_view name{ "config::bus" };
  };
};
```

magni-mar:
Naming conventions can be reviewed.

omarhogni (changes requested):
Missing removal of --iface parameter in systemd template also.

exes/ethercat/inc/public/tfc/ec.hpp
```diff
@@ -371,6 +376,8 @@ class context_t {
   int32_t wkc_ = 0;
   std::array<std::byte, pdo_buffer_size> io_;
   std::unique_ptr<std::thread> check_thread_;
+  tfc::confman::config<config::bus> config_{ ctx_, "ethercat" };
```
omarhogni, line 379: Could you use the dbus_ connection we added to save connections here?

magni-mar: Yes, it's possible but the confman constructor needs to be changed. Also the config needs to be changed to a shared_ptr and moved, please see changes.

libs/confman/inc/public/tfc/confman.hpp
```diff
@@ -77,7 +77,7 @@ class config {
   config(std::shared_ptr<sdbusplus::asio::connection> conn, std::string_view key, storage_type&& def)
       : client_{ conn, key, std::bind_front(&config::string, this), std::bind_front(&config::schema, this),
                  std::bind_front(&config::from_string, this) },
-        storage_{ client_.io_context(), tfc::base::make_config_file_name(key, "json"), std::forward<storage_type>(def) },
+        storage_{ client_.get_io_context(), tfc::base::make_config_file_name(key, "json"), std::forward<storage_type>(def) },
```
omarhogni, line 80: @jbbjarnason Can you comment on if and why this is needed. Everything still seems to build so maybe this constructor has not been used project wide.

jbbjarnason: this constructor is most likely missing a test of usage

exes/ethercat/inc/public/tfc/ec.hpp
```diff
@@ -151,10 +149,21 @@ class context_t {
...
+      // TODO: switch for error_code
+      throw std::runtime_error("Failed to init, no socket connection");
+    }
+
     if (!config_init(false)) {
-      // TODO: Switch for error_code
-      throw std::runtime_error("No slaves found!");
+      /// Since the network interface is in a config file which only lives as long as the program is running the user needs
+      /// time to configure another interface if this one fails
+      logger_.error("No slaves found for interface: {}", config_->value().primary_interface.value);
```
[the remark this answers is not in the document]
magni-mar: why?

exes/ethercat/inc/public/tfc/ec.hpp
```diff
@@ -151,10 +149,21 @@ class context_t {
...
+      logger_.error("Might need to configure another interface?");
+      ctx_.run_for(std::chrono::seconds{ 1 });
```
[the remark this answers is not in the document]
magni-mar: then you couldn't return the error code from the function, or the error code would be returned as something other than the final code actually is

exes/ethercat/inc/public/tfc/ec/config/ethercat.hpp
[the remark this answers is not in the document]
magni-mar: I prefer `tfc/ec/config/ethercat.hpp` because in other places, such as mqtt bridge, this file convention is maintained.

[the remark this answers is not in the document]
jbbjarnason: hmm okay, `ethercat.hpp` is a bit too general don't you think? Would it be appropriate to name this file `bus.hpp`

exes/ethercat/inc/public/tfc/ec/config/ethercat.hpp
```diff
@@ -0,0 +1,57 @@
...
+template <>
+struct tfc::json::detail::to_json_schema<tfc::ec::config::network_interface> {
+  template <auto Opts>
+  static void op(auto& s, auto&) noexcept {
+    s.oneOf = std::vector<tfc::json::detail::schematic>{};
+
+    // clang-format off
+        PRAGMA_CLANG_WARNING_PUSH_OFF(-Wexit-time-destructors)
+        PRAGMA_CLANG_WARNING_PUSH_OFF(-Wglobal-constructors)
+        thread_local auto interfaces{tfc::global::get_interfaces()};
```
jbbjarnason, line 36: maybe it was not elaborated well enough in previous comment. 

In relation to set_interfaces comment, the before mentioned function can be implemented in the source file and store the thread local instance there.
However it does not need to be predeclared in header as a user does not need to call set_interfaces.

What I am trying to say, with reference to `interfaces.cpp`:

```cpp
namespace {
thread_local std::vector<std::string> interfaces{};
}
namespace tfc::global {

auto get_interfaces() -> std::vector<std::string> const& { // or std::span<std::string> or better std::span<std::string_view> if it works
  struct ifaddrs* addrs;
  getifaddrs(&addrs);

  interfaces.clear();

  for (struct ifaddrs* addr = addrs; addr != nullptr; addr = addr->ifa_next) {
    if (addr->ifa_addr && addr->ifa_addr->sa_family == AF_PACKET) {
      interfaces.emplace_back(addr->ifa_name);
    }
  }
  freeifaddrs(addrs);

  return interfaces;
}

}  // namespace tfc::global
```

jbbjarnason (approved):

PR #363, opened 2024-01-11
[mqtt_bridge] Minimize dropdown for mqtt bridge

magni-mar:


magni-mar:
Waiting with this until a proper solution will be found in accordance with the frontend.

omarhogni:
same result as #395

PR #366, opened 2024-01-11
[ethercat] ensure a valid interface is chosen for ethercat

magni-mar:


exes/ethercat/inc/public/tfc/ec.hpp
```diff
@@ -65,9 +66,22 @@ class context_t {
...
 
+  auto is_interface_valid() -> bool {
+    for (auto const& interface : tfc::global::get_interfaces()) {
+      if (interface == config_->value().primary_interface.value) {
+        return true;
+      }
+    }
+    return false;
+  }
+
```
omarhogni, line 84: Please use std::find or std::ranges::find

omarhogni (changes requested):

omarhogni (approved):
Lgtm

PR #379, opened 2024-01-17
Added missing pragma once to motor.hpp

omarhogni:


jbbjarnason (approved):

omarhogni:
@jbbjarnason There is now a github action check that searches all .hpp files under lib and exes to see if there are missing pragmas.

magni-mar (approved):
lgtm

PR #380, opened 2024-01-17
Reduce file descriptors

jbbjarnason:
Key retained:

image

Secondly added source location support:

image

### DIFF
--- a/libs/logger/inc/custom_sink.hpp
+++ b/libs/logger/inc/custom_sink.hpp
@@ -65,12 +65,6 @@ protected:
       payload = msg.payload;
     }
 
-    size_t length = payload.size();
-    // limit to max int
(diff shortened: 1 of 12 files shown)


libs/logger/inc/custom_sink.hpp
```diff
@@ -65,14 +64,8 @@ class tfc_systemd_sink : public base_sink<mutex> {
       payload = msg.payload;
     }
 
-    size_t length = payload.size();
-    // limit to max int
-    if (length > static_cast<size_t>(std::numeric_limits<int>::max())) {
-      length = static_cast<size_t>(std::numeric_limits<int>::max());
-    }
-
```
omarhogni, line 73: I wonder what the reason was for this. I am making this note so that I can remember to check the max message size.

omarhogni (approved):
Very good changes here

PR #382, opened 2024-01-18
Fix optional access in move home

jbbjarnason:


exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
```diff
@@ -584,9 +584,10 @@ struct controller {
             }
             case state_e::wait_till_stop: {
               state = state_e::complete;
-              // let's verify that we are not positioned at home
-              // if we are not in sensor, check whether we are cancelled
-              if (!homing_sensor->value().value() && err == std::errc::operation_canceled) {
+              if (homing_sensor->value().value_or(false)) {
+                // let's continue to stopping the conveyor since the value of home sensor is high
+              }
```
jbbjarnason: we should remove this code when limit switch is implemented

omarhogni (approved):

PR #392, opened 2024-01-23
Sorted config items in sidebar

### LINKED ISSUE #354: [cockpit] ascending/descending order of items
Would be nice to have elements in cockpit sorted by ascending/descending numbers, letters and such in the Configurator.

image

SindriTh:
Items in sidebar are now sorted.

image

closes #354

### DIFF
--- a/cockpit/src/views/Configurator.tsx
+++ b/cockpit/src/views/Configurator.tsx
@@ -51,6 +51,24 @@ const Configurator: React.FC = () => {
         ) && !name.includes('ipc_ruler'),
       );
 
+      filteredNames.sort((a, b) => {
+        const regex = /([^\d]+)(\d+)$/;
+        const matchA = regex.exec(a);
+        const matchB = regex.exec(b);
+
+        if (matchA && matchB) {
+          // Compare non-numeric parts
+          if (matchA[1] < matchB[1]) return -1;
+          if (matchA[1] > matchB[1]) return 1;
+
+          // Compare numeric parts
+          return parseInt(matchA[2], 10) - parseInt(matchB[2], 10);
+        }
+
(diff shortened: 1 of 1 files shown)


cockpit/IPC/src/views/Configurator.tsx
```diff
@@ -50,8 +50,26 @@ const Configurator: React.FC = () => {
         ) && !name.includes('ipc_ruler'),
       );
 
+      const sortedNames = filteredNames.sort((a, b) => {
+        const regex = /(\D+)(\d+)$/;
+        const matchA = a.match(regex);
+        const matchB = b.match(regex);
+
+        if (matchA && matchB) {
```
omarhogni, line 58: Should we not check the length here just to be sure?

omarhogni (changes requested):

cockpit/IPC/src/views/Configurator.tsx
SindriTh: > Should we not check the length here just to be sure?

`matchA` and `matchB` will either be null (if no match is found) or an array where:

- matchA[0] or matchB[0] is the full string that matches the regex.
- matchA[1] or matchB[1] is the first captured group (non-numeric part).
- matchA[2] or matchB[2] is the second captured group (numeric part).

The condition if (matchA && matchB) is checking if both matchA and matchB are not null, which means both strings matched the regex pattern. If this condition is true, then it is guaranteed that both matchA and matchB are arrays with at least three elements (the entire match, the non-numeric part, and the numeric part). Therefore, it's safe to access matchA[1], matchA[2], matchB[1], and matchB[2] without an additional length check.

omarhogni (approved):

PR #411, opened 2024-01-31
Change slot cb to weak ptr

magni-mar:
ZeroMQ won't destruct if there are still shared pointers which haven't been destroyed, this problem can be circumvented by changing the storing of slots to weak ptrs.


omarhogni:
@jbbjarnason we added this as a means to gracefully deconstruct mqtt_bridge. having references to shared_ptr's which hold zeromq context does not deconstruct.

jbbjarnason:
> @jbbjarnason we added this as a means to gracefully deconstruct mqtt_bridge. having references to shared_ptr's which hold zeromq context does not deconstruct.

@omarhogni who holds the shared pointer and why is the owner still keeping reference, zeromq does not keep any reference so I do not see how this fixes something. Does this change make sense to you? Is the slot generated elsewhere and the variant is used to point to that slot or how does this work?

omarhogni:
This was a co-routine issue in the end. Changes not needed.

PR #419, opened 2024-02-02
Limit switch

jbbjarnason:
I am not sure how to handle the case when home sensor is the same as limit switch, it could be random whether you would get a limit error during move_home or success

### DIFF
--- a/exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
+++ b/exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
@@ -629,9 +629,7 @@ private:
             }
             case state_e::wait_till_stop: {
               state = state_e::complete;
-              if (homing_sensor->value().value_or(false)) {
-                // let's continue to stopping the conveyor since the value of home sensor is high
-              } else if (err == std::errc::operation_canceled) {
+              if (err == std::errc::operation_canceled) {
                 logger_.trace("Move home got cancelled");
                 self.complete(err);
                 return;
--- a/exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
+++ b/exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
@@ -661,8 +661,6 @@ private:
   tfc::asio::condition_variable<asio::any_io_executor> drive_error_subscriptable_{ ctx_.get_executor() };
   tfc::asio::condition_variable<asio::any_io_executor> homing_complete_{ ctx_.get_executor() };
   asio::cancellation_signal cancel_signal_{};
-  asio::cancellation_signal positive_limit_signal_{};
-  asio::cancellation_signal negative_limit_signal_{};
   asio::cancellation_signal no_drive_error_{};
   logger::logger logger_{ fmt::format("{}_{}", impl_name, slave_id_) };
 
--- a/exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
+++ b/exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
@@ -65,7 +65,8 @@ combine_error_code(completion_token_t&&) -> combine_error_code<completion_token_
 template <typename completion_token_t>
 struct drive_error_first {
   drive_error_first(completion_token_t&& token, motor::errors::err_enum& drive_error, motor::errors::err_enum& limit_error)
-      : drive_err{ drive_error }, limit_err{ limit_error}, self{ std::move(token) }, slot{ asio::get_associated_cancellation_slot(self) } {}
+      : drive_err{ drive_error }, limit_err{ limit_error }, self{ std::move(token) },
+        slot{ asio::get_associated_cancellation_slot(self) } {}
 
   /// The associated cancellation slot type.
   using cancellation_slot_type = asio::cancellation_slot;
@@ -300,8 +301,7 @@ private:
     if (new_v) {
       limit_error_ = limit_error;
       drive_error_subscriptable_.notify_all();
-    }
-    else if (limit_error_ == limit_error) {
+    } else if (limit_error_ == limit_error) {
       limit_error_ = success;
     }
   }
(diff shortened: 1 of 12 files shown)


exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
```diff
@@ -276,8 +296,18 @@ struct controller {
...
 
 private:
+  auto on_limit_switch(bool new_v, motor::errors::err_enum limit_error) {
+    using enum motor::errors::err_enum;
+    if (new_v) {
+      limit_error_ = limit_error;
+      drive_error_subscriptable_.notify_all();
+    } else if (limit_error_ == limit_error) {
+      limit_error_ = success;
+    }
```
omarhogni, line 306: Could we add some comments around this logic. It does not read well.

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
```diff
@@ -314,6 +345,17 @@ struct controller {
         token);
   }
 
+  auto run_limitted(bool positive_speedratio) -> bool {
```
omarhogni, line 348:
```suggestion
  auto run_limited(bool positive_speedratio) -> bool {
```
could also be `run_from_limit` or `run_outoff_limit`

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
```diff
@@ -314,6 +345,17 @@ struct controller {
...
+    using enum motor::errors::err_enum;
+    if (limit_error_ != success) {
```
omarhogni, line 350: Does this matter?

omarhogni: I see now this method is inversed, maybe we should call it `forbid_run_limited` then a true value indicates that running is forbidden.

omarhogni (commented):
I think it is crucial for us to deal with this race condition before we can setup a scenario where the homing sensor and one of the limit switches are the same. 

However I am not really sure how we can do that without checking that one off the limit switches is connected to the same signal as the homing sensor. Which would require us the expose internals not currently exposed.

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
jbbjarnason: Hmm yes the name is neither obvious or good. How about `is_forbidden`?

jbbjarnason:
> I think it is crucial for us to deal with this race condition before we can setup a scenario where the homing sensor and one of the limit switches are the same. 
> 
> However I am not really sure how we can do that without checking that one off the limit switches is connected to the same signal as the homing sensor. Which would require us the expose internals not currently exposed.

That's pretty good solution for that problem!
It's trivial to expose it's connection and the information is contained in same code path.

Should it be implemented here or as a follow-up?

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
omarhogni: `is_forbidden` is good.

omarhogni:
> Should it be implemented here or as a follow-up?
Both fine by me.

libs/motor/inc/public/tfc/motor/details/positioner_impl.hpp
```diff
@@ -49,6 +49,8 @@ using position_mode_config = std::variant<std::monostate,
                                           encoder_config<reference>,
                                           freq_config<deduce_velocity_t<reference>>>;
 
+enum struct limit_switch_e : std::uint8_t { not_used, limit_switches_optional };
```
omarhogni, line 52: I don't quite get the point of having this.

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
```diff
@@ -248,10 +257,33 @@ struct controller {
...
+  void on_positive_limit_switch(bool new_v) {
+    positive_limit_value_ = new_v;
+    logger_.trace("New positive limit switch value: {}", new_v);
+    if (pos_.positive_limit_switch().has_value() && pos_.homing_sensor().has_value()) {
+      if (pos_.positive_limit_switch().value().connection() == pos_.homing_sensor().value().connection()) {
+        homing_complete_.notify_all();
+      }
+    }
+    on_limit_switch(new_v, motor::errors::err_enum::positioning_positive_limit_reached);
+  }
```
omarhogni, line 269: I am kind of surprised this works. I would think `on_limit_switch` would be called before the callback of `notify_all` could it be this code is never "hit" in testing cause the homing_sensor connects first to the signal or do you have more insight here?

omarhogni (approved):

libs/motor/inc/public/tfc/motor/details/positioner_impl.hpp
jbbjarnason: Do you mean on a config level or whether this should be a bool ?

exes/ethercat/inc/public/tfc/ec/devices/schneider/atv320/dbus-iface.hpp
omarhogni: my thought was that `on_limit_switch` is not a async initiator but executes its logic as soon as called. but notify_all probably just schedules work onto an execution context.

I would have thought there should be a return after `homing_complete_.notify_all();`

jbbjarnason: First paragraph answer: on_limit_switch is also scheduling work onto the execution context, therefore, it will initiate after the homing_complete_ token is scheduled.

Second paragraph answer: Yes I skipped the return for a reason, if someone calls `move(0)` it should stop at the limit switch even though it is connected to the same input as homing sensor. In that call stack no one is listening to homing_complete_ condition so that one will not activate any token. 


Is this logic too complex or do you have suggestions on making this more clear?

PR #422, opened 2024-02-02
separate strategy between tachometer and encoder to detect missing event

jbbjarnason:


libs/motor/inc/public/tfc/motor/details/positioner_impl.hpp
```diff
@@ -226,16 +223,22 @@ struct encoder {
...
+    errors::err_enum err{ errors::err_enum::success };
     if (buffer_.front().last_event == event) {
-      err = errors::err_enum::positioning_missing_event;
+      // if positioning is still going to the same direction, we should have gotten a new event
+      // needs revisement if increment is dynamic, as in not always 1 or -1
+      if (increment == last_increment_) {
+        err = errors::err_enum::positioning_missing_event;
+      }
     }
+    last_increment_ = increment;
```
omarhogni, line 234: Can this ever work? increment needs to be zero for this to trigger?

libs/motor/inc/public/tfc/motor/details/positioner_impl.hpp
```diff
@@ -168,8 +159,14 @@ struct tachometer {
     statistics_.update(now);
     auto constexpr increment{ 1 };
     position_ += increment;
-    std::invoke(position_update_callback_, increment, statistics().average(), statistics().stddev(),
-                detect_deviation_from_average(statistics_.last_interval(), statistics_.average()));
+    auto err{ errors::err_enum::success };
+    // check if interval is greater than 180% of average and less than 400% of average
+    // todo is 400% good strategy to detect when stopped and started again?
+    if (statistics_.last_interval() * 10 > statistics_.average() * 18 &&
+        statistics_.last_interval() < statistics_.average() * 4) {
```
omarhogni, line 166: This is confusing, where is the original 400% number coming from, what are we trying to avoid with this change. Could we express this in more clear terms.

omarhogni (changes requested):
I need am not convinced these are the right changes. Can we reproduce the errors occuring on a physical level with the unit tests for positioner?

I think it is crucial for changes like these where the modification is so subtle and not at all obvious to have a failing test along with them in order to avoid regression.

We can also discuss this in person and go over the ideas more holistically

libs/motor/inc/public/tfc/motor/details/positioner_impl.hpp
jbbjarnason: 400% was decided based on experience, normally you could miss 2 - 3 pulses or up to 300 % +

but this strategy can be discussed, this at least is more stable than it was before. It is really weird to report missing event if motor is starting.

omarhogni (approved):

omarhogni (approved):

PR #428, opened 2024-02-06
integration test mqtt

magni-mar:


magni-mar:
the code fails with a dwarf2out error when building integration tests, unknown reason.

magni-mar:
ICE is solved by #469

magni-mar:
disabled compiling tests on gcc, to many updates to container need to be done in order for that to be viable. However, should be fine locally if gcc is at a late enough version.

exes/mqtt-bridge/inc/external_to_tfc.hpp
```diff
@@ -2,64 +2,93 @@
...
+
+  auto receive_new_value(std::string signal_name, std::variant<bool, double, std::string, int64_t, uint64_t> value) -> void {
+    logger_.trace("Received new value for signal: {}", signal_name);
+
+    std::visit(
+        [&value]<typename signal_t>(signal_t& signal) {
+          if constexpr (!std::is_same_v<std::remove_cvref_t<signal_t>, std::monostate>) {
+            using value_t = typename std::remove_cvref_t<signal_t>::value_t;
+
+            // Spark Plug B sends int64 values as uint64
```
omarhogni, line 53:
```suggestion
            // Spark Plug B sends uint64 values as int64
```

exes/mqtt-bridge/inc/external_to_tfc.hpp
```diff
@@ -2,64 +2,93 @@
...
+            // when an int64 value arrives it is in the form of uint64
+            // therefore the uint64 value needs to be "get" and cast into int64
+            // to the signal
+            if constexpr (std::is_same_v<value_t, int64_t>) {
+              signal.send(static_cast<int64_t>(std::get<uint64_t>(value)));
+            } else if constexpr (tfc::stx::is_expected_quantity<value_t>) {
+              // todo
```
omarhogni, line 60: maybe add a log here , unlikely some client will try to write these values and it is unexpected.

exes/mqtt-bridge/inc/external_to_tfc.hpp
```diff
@@ -2,64 +2,93 @@
...
+    std::string segment;
+    std::vector<std::string> seglist;
+
+    while (std::getline(test, segment, '/')) {
+      seglist.push_back(segment);
+    }
+
+    if (!seglist.empty()) {
+      return seglist.back();
+    }
```
omarhogni, line 80: could reduce memory here by using a std::optional<std::string> last_segment and ditch the vector. Can you not do this simply with std::ranges?

you could also simplify using std::find_last `/` if that is the end of the string, none is found and then return a sub-string from that index to the end.

exes/mqtt-bridge/src/spark_plug_interface.cpp
```diff
@@ -18,21 +18,23 @@
 #include <spark_plug_interface.hpp>
 #include <structs.hpp>
 
+// #include "../inc/endpoint_mock.hpp"
+
```
omarhogni, line 22:
```suggestion
```

omarhogni (changes requested):

exes/mqtt-bridge/inc/external_to_tfc.hpp
magni-mar: simplified this function with std::ranges as you pointed out and added tests for it

exes/mqtt-bridge/inc/external_to_tfc.hpp
magni-mar: This is an incorrect suggestion, Spark Plug B sends the value as uint64 and the value is cast into int64.

vcpkg.json
```diff
@@ -13,6 +13,7 @@
     "boost-program-options",
     "boost-coroutine2",
     "boost-stacktrace",
+    "boost-uuid",
```
jbbjarnason, line 16: where is this used?

magni-mar: this was used, but when building without this everything still builds, removing this.

exes/mqtt-bridge/inc/client.hpp
```diff
@@ -1,78 +1,235 @@
...
+    auto pub_packet = async_mqtt::v5::publish_packet{ p_id.value(), async_mqtt::allocate_buffer(topic),
+                                                      async_mqtt::allocate_buffer(payload), qos };
 
-  auto send_initial() -> asio::awaitable<bool>;
+    auto send_error = co_await endpoint_client_->send(pub_packet, asio::use_awaitable);
+
+    if (send_error) {
+      co_return false;
+    }
+    co_return true;
```
jbbjarnason, line 113:
```suggestion
    co_return !(co_await endpoint_client_->send(pub_packet, asio::use_awaitable));
```
Does this work?

exes/mqtt-bridge/inc/spark_plug_interface.hpp
```diff
@@ -52,69 +32,278 @@ using org::eclipse::tahu::protobuf::Payload_Metric;
...
+  auto make_will_payload() -> std::string {
+    /// Construct MQTT will payload
+    Payload payload;
+    payload.set_timestamp(timestamp_milliseconds().count());
+
+    auto* metric = payload.add_metrics();
+
+    metric->set_name("bdSeq");
+    metric->set_timestamp(timestamp_milliseconds().count());
+    metric->set_datatype(8);
```
jbbjarnason, line 57: what does 8 mean?

exes/mqtt-bridge/inc/spark_plug_interface.hpp
```diff
@@ -52,69 +32,278 @@ using org::eclipse::tahu::protobuf::Payload_Metric;
...
+    if (mqtt_client_) {
+      Payload payload;
+      payload.set_timestamp(timestamp_milliseconds().count());
+      seq_ = 0;
+      payload.set_seq(seq_);
+
+      auto* node_rebirth = payload.add_metrics();
+      node_rebirth->set_name(constants::rebirth_metric.data());
+      node_rebirth->set_timestamp(timestamp_milliseconds().count());
+      node_rebirth->set_datatype(11);
```
jbbjarnason, line 88: same

exes/mqtt-bridge/inc/spark_plug_interface.hpp
```diff
@@ -52,69 +32,278 @@ using org::eclipse::tahu::protobuf::Payload_Metric;
...
+      node_rebirth->set_boolean_value(false);
+      node_rebirth->set_is_transient(false);
+      node_rebirth->set_is_historical(false);
+      node_rebirth->set_is_null(false);
+
+      auto* bd_seq_metric = payload.add_metrics();
+
+      bd_seq_metric->set_name("bdSeq");
+      bd_seq_metric->set_timestamp(timestamp_milliseconds().count());
+      bd_seq_metric->set_datatype(8);
```
jbbjarnason, line 98: same

exes/mqtt-bridge/inc/spark_plug_interface.hpp
magni-mar: 8 is uint64 data type, I can add comment to clarify

exes/mqtt-bridge/inc/spark_plug_interface.hpp
magni-mar: 11 is boolean

exes/mqtt-bridge/inc/spark_plug_interface.hpp
jbbjarnason: does there not exist an enum, which can be used?

magni-mar: yes, updating to enum

jbbjarnason (approved):

magni-mar:
pipeline won't pass until #515 is merged

Issue #175, opened 2023-10-04
MQTT bridge cannot dynamically load signals


magni-mar:
The MQTT bridge, pending #153, cannot dynamically load new signals that are created. It must be restarted in order for the signals to arrive.

magni-mar:
In my opinion the MQTT bridge should publish all signals and then a blacklist should be implemented. When either a new signal is registered, or a new signal is added to the blacklist the program reloads which signals it is publishing.

Thoughts? @omarhogni  @jbbjarnason

PR #429, opened 2024-02-06
mqtt dynamic signals

magni-mar:
#428 first
#175 issue

load signals dynamically, tests are included.


exes/mqtt-bridge/inc/run.hpp
```diff
@@ -27,64 +30,58 @@ class run {
...
-        co_await asio::steady_timer{ sp_interface_.strand(), std::chrono::seconds{ 5 } }.async_wait(asio::use_awaitable);
+      while (!restart_needed_) {
+        co_await asio::steady_timer{ sp_interface_->strand(), std::chrono::milliseconds{ 100 } }.async_wait(
+            asio::use_awaitable);
       }
 
       cancel_signal.emit(asio::cancellation_type::all);
-      tfc_to_exter_.clear_signals();
+
+      io_ctx_.run_for(std::chrono::milliseconds{ 10 });
```
jbbjarnason, line 70: why delay?

jbbjarnason (commented):
I am not fond of restart_needed polling, could it be made more event based instead of references and polling?

exes/mqtt-bridge/inc/run.hpp
magni-mar: allow sockets to clear up and connections to close, if this is skipped then the next run will fail

magni-mar:
> I am not fond of restart_needed polling, could it be made more event based instead of references and polling?

I am open to suggestions

exes/mqtt-bridge/inc/run.hpp
jbbjarnason: I see, can you add that to a comment above?

exes/mqtt-bridge/inc/run.hpp
```diff
@@ -27,64 +30,58 @@ class run {
...
-          bind_cancellation_slot(cancel_signal.slot(), asio::detached));
-
-      io_ctx_.run_for(std::chrono::seconds{ 1 });
-
-      while (!restart_needed) {
-        co_await asio::steady_timer{ sp_interface_.strand(), std::chrono::seconds{ 5 } }.async_wait(asio::use_awaitable);
+      while (!restart_needed_) {
+        co_await asio::steady_timer{ sp_interface_->strand(), std::chrono::milliseconds{ 100 } }.async_wait(
+            asio::use_awaitable);
       }
```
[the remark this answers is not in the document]
magni-mar: Looking into this and will ping this thread when I have a solution, are you talking about `boost::asio::completion_token`?

exes/mqtt-bridge/inc/run.hpp
magni-mar: yes no problem

exes/mqtt-bridge/inc/run.hpp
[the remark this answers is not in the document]
magni-mar: This can be implemented for reloading the application based on new signals, there is another reloading which happens when an incorrect packet is received. In my experience, when that happens the program goes into an endless loop and needs to be restarted anyways so I removed that. If that becomes a problem in the future it needs to be investigated on its own.

exes/mqtt-bridge/tests/src/integration_tests.cpp
```diff
@@ -219,46 +220,222 @@ auto main(int argc, char* argv[]) -> int {
...
+    //      timestamp: 1706985768834
+    //      datatype: 11
+    //      is_historical: false
+    //      is_transient: false
+    //      is_null: true
+    //      metadata {
+    //          description: ""
+    //        }
+    //  }
+    //  seq: 0
```
omarhogni, line 345: ?

libs/motor/inc/public/tfc/motor/atv320motor.hpp
omarhogni, line 1: rebase please

exes/mqtt-bridge/inc/config/bridge.hpp
omarhogni, line 1: would have been nice to do this in a separate PR

exes/mqtt-bridge/inc/tfc_to_external.hpp
```diff
@@ -78,19 +93,18 @@ class tfc_to_external {
 
   auto is_publish_signal(std::string signal_name) -> bool {
     // check if signal is a writeable signal
+    for (const auto& banned_signal : config_.value().banned_signals) {
+      if (banned_signal.value == signal_name) {
+        logger_.trace("Signal {} is in the list of banned signals", signal_name);
```
omarhogni, line 98: This can be higher than trace. info would be proper.

omarhogni (changes requested):
there are no tests for new functionality banned_signals.

exes/mqtt-bridge/inc/config/bridge.hpp
magni-mar: this relates to dynamic loading of signals so that they are discovered when the program is reloaded with new signals, otherwise the config blocks them

magni-mar:
> there are no tests for new functionality banned_signals.

the last test in `integration_tests` is a test for banned_signals

jbbjarnason (commented):
This change makes it more easily configurable from my understanding, however, there is a small change which has a rather big impact.

The change of config, from being a whitelist(publish_signals) to a blacklist(banned_signals) this will make it harder to update at customer site. Why this change of course?

What I thought to be the future of this component/interface would be like an option in the IO connections view to enable a signal being published with a simple checkmark or something. I think a whitelist is better for that purpose.

@omarhogni I think we need to make a documentation in regards to future of this components interface and functionality based on our first trial.
Maybe we could use Github Projects for it, if it is beneficial.

magni-mar:
> This change makes it more easily configurable from my understanding, however, there is a small change which has a rather big impact.
> 
> The change of config, from being a whitelist(publish_signals) to a blacklist(banned_signals) this will make it harder to update at customer site. Why this change of course?
> 
> What I thought to be the future of this component/interface would be like an option in the IO connections view to enable a signal being published with a simple checkmark or something. I think a whitelist is better for that purpose.
> 
> @omarhogni I think we need to make a documentation in regards to future of this components interface and functionality based on our first trial. Maybe we could use Github Projects for it, if it is beneficial.

Regardless of weather this function is implemented or not, there are invaluable tests in this PR that should be finished and merged. I can change this PR to tests only and implement dynamic signals. Thoughts?

jbbjarnason:
Let's reopen if we need this feature

PR #431, opened 2024-02-07
Mass signal in MQTT bridge using templates

magni-mar:
This is a stacked PR upon these:  #428  #429 . So when those are merged some of these files will disappear.

AVEVA seemingly doesn't support templating types, however, more information on this subject might be necessary to fully conclude this.


magni-mar:
> This is a stacked PR upon these: #428 #429 . So when those are merged some of these files will disappear.
> 
> AVEVA seemingly doesn't support templating types, however, more information on this subject might be necessary to fully conclude this.

AVEVA does support this with very specific parameters for topic names, implemented but yet to be tested.

magni-mar:
AVEVA doesn´t understand the NBIRTH and provides no error logs to debug why.

jbbjarnason:
@magni-mar should we change this to an issue and reference this PR? Seems like it does not work on Aveva side?

magni-mar:
> @magni-mar should we change this to an issue and reference this PR? Seems like it does not work on Aveva side?

Sounds good, I don't see the need for this as of now.

jbbjarnason:
Changed to an issue

PR #432, opened 2024-02-07
Added cancel behaviour to stub to emulate ATV320 behaviour

omarhogni:


jbbjarnason (approved):
I have mixed feelings about this, since this logic is currently not generalized for all motor implementations. But I would like it being so, not sure how it would be best implemented.

jbbjarnason:
You could refactor similar to motor.hpp to reduce duplication

omarhogni:
Will merge for now can discuss later and make changes.

PR #441, opened 2024-02-09
add cloudsmith cli to container

magni-mar:


magni-mar:
Unnecessary to add this to the container as this is seldomly used, better to pip install cloudsmith into a venv and use it from there. Closing this PR.

PR #445, opened 2024-02-09
mock operation mode

omarhogni:
add a type which adheres to operations interface class in order to inject the type and mock functionality for testing.


libs/operation_mode/testing/mocks/inc/public/tfc/mocks/operation_mode.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include <gmock/gmock.h>
+#include <boost/asio/io_context.hpp>
+
+#include <tfc/operation_mode.hpp>
+
+namespace tfc::operation {
+
+namespace asio = boost::asio;
+
+struct mock_interface {
```
jbbjarnason, line 15: how about stub_interface ?

jbbjarnason (approved):

PR #458, opened 2024-02-13
fix motor config

jbbjarnason:
Fix null interpretation, std::monostate inside std::shared ptr does not work well. Interpretes std::monostate as std::shared_ptr when reading json

Secondly, use tagged variant for the future. Can support multiple types of frequency drives with same config parameters, i.e. slave_id.

### DIFF
--- a/libs/motor/inc/public/tfc/motor/stub.hpp
+++ b/libs/motor/inc/public/tfc/motor/stub.hpp
@@ -36,6 +36,7 @@ struct stub {
     struct glaze {
       using T = config;
       static constexpr auto value = glz::object("name", &T::name);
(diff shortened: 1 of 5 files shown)


omarhogni (approved):

libs/motor/inc/public/tfc/motor/stub.hpp
```diff
@@ -36,6 +36,7 @@ struct stub {
     struct glaze {
       using T = config;
       static constexpr auto value = glz::object("name", &T::name);
+      // Todo use tfc::json::schema meta and expose variant without this type to UI
```
jbbjarnason, line 39: do this

PR #460, opened 2024-02-13
Fix confman mocks to install

omarhogni:


libs/confman/testing/CMakeLists.txt
```diff
@@ -3,4 +3,6 @@ if(BUILD_EXAMPLES)
 endif()
 add_subdirectory(mocks)
 add_subdirectory(stubs)
-add_subdirectory(tests)
+if(BUILD_TESTS)
```
jbbjarnason, line 6:
```suggestion
if(BUILD_TESTING)
```
where is build_tests from?

jbbjarnason (approved):

PR #469, opened 2024-02-13
Upgrade base image of container to ubuntu 23:10

magni-mar:
#428  had an unexplainable ICE error when building and it was running gcc version 13.1.0-2ubuntu2~23.04 , this updated container is running 13.2.0-4ubuntu3 and the branch can build. All build targets were tested with gcc-debug locally and they could build.

### DIFF
--- a/containers/tfc-toolchain/tfc-toolchain.dockerfile
+++ b/containers/tfc-toolchain/tfc-toolchain.dockerfile
@@ -1,5 +1,5 @@
 # boost-build requires newer compiler than gcc 8
-FROM ubuntu:23.04 AS base
+FROM ubuntu:23.10 AS base
 
 ENV DEBIAN_FRONTEND noninteractive
 


jbbjarnason:
How about only uprading the compiler version in 23.04 not the standard library nor libc?

There is probably some upstream testing mirror which can be enabled - install gcc - disabled

That would downsize the impact of this change.

magni-mar:
Will look into a more targeted install of latest gcc

jbbjarnason:
I think we need to stick to an older version of glibc, the version available on debian testing is 2.37 https://packages.debian.org/sid/libc6 where ubuntu 23.10 is 
```bash
ldd --version
ldd (Ubuntu GLIBC 2.38-1ubuntu6) 2.38
```

magni-mar:
closing this as the ICE was solved in another manner

PR #478, opened 2024-02-14
check size of output before indexing it

jbbjarnason:


omarhogni (approved):
A single if with an early return is enough. 

This is called so often. Would maybe be benoficial to log this error on first occurance.

omarhogni (approved):

exes/ethercat/inc/tfc/ec/devices/beckhoff/EL2xxx_impl.hpp
```diff
@@ -29,6 +29,14 @@ template <typename manager_client_type,
           tfc::stx::basic_fixed_string name>
 void el2xxx<manager_client_type, size, entries, pc, name>::process_data(std::span<std::byte>,
                                                                         std::span<std::byte> output) noexcept {
+  if (output.size() == 0) {
+    if (output_buffer_valid_) {
+      logger_.info("No output buffer provided for EL2xxx {}", name);
```
jbbjarnason, line 34:
```suggestion
      logger_.info("No output buffer provided for EL2xxx {}", name.view());
```

PR #489, opened 2024-02-16
Fixed negative number handling on UnitWidget

### LINKED ISSUE #486: Unable to configure numbers to minus
example speedratio

SindriTh:
fixes #486 

Changed input handling regex to handle negative numbers.

This change also fixes a bug where you couldn't place a decimal point in the middle of an existing string.


jbbjarnason:
tested and works on target

jbbjarnason (approved):

PR #520, opened 2024-02-28
docs function

magni-mar:


magni-mar:
https://github.com/Skaginn3x/cmake_modules/pull/15 needs to merge first, then the submodule path updated

jbbjarnason (approved):

PR #524, opened 2024-03-21
update toolchain to llvm 18

jbbjarnason:
I would like to reinterpret the buffer before calling process_data, what's your opinion on that @omarhogni ?

### DIFF
--- a/containers/tfc-toolchain/tfc-toolchain.dockerfile
+++ b/containers/tfc-toolchain/tfc-toolchain.dockerfile
@@ -73,11 +73,11 @@ RUN apt clean all
 RUN rm -rf /var/lib/apt/lists/*
 RUN rm -rf /tmp/*
 
-FROM gcc-13 as clang-17
+FROM gcc-13 as clang-18
 
 COPY shared.sh /tmp/
 COPY build-clang.sh /tmp/
-RUN ./build-clang.sh 17.0.6
+RUN ./build-clang.sh 18.1.2
 
 #ENV CC=clang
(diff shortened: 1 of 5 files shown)


omarhogni:
> I would like to reinterpret the buffer before calling process_data, what's your opinion on that @omarhogni ?

to which type?

omarhogni (approved):

arturbac:
@jbbjarnason You are missing alignment check for reinterpret cast, on x86_64 it will work unaligned but will be slow, on some older arms it will just crash, on modern >= armv7 unaligned should work but same as on x64 very slow
And I would use some span_cast as below to avoid mistakes with size

```
template<typename T, typename... Args>
concept same_as_any_of = std::disjunction_v<std::is_same<T, Args>...>;

template<typename T>
concept byte_like = same_as_any_of<T, std::byte, char, uint8_t>;

template<typename T>
concept trivially_copyable_non_byte = std::is_trivially_copyable_v<T> && !byte_like<T>;

template<byte_like SrcType, trivially_copyable_non_byte DstType>
[[nodiscard]] 
constexpr auto span_cast(std::span<SrcType> src) -> std::span<const DstType> {
    assert((reinterpret_cast<std::uintptr_t>(src.data()) % alignof(DstType) == 0) && "Data must be aligned to the destination type");
    std::size_t dst_size = src.size_bytes() / sizeof(DstType);
    return {reinterpret_cast<const DstType*>(src.data()), dst_size};
}
```

Issue #513, opened 2024-02-22
Operations initial values on signals


jbbjarnason:
In constructor set signals to false or empty string, other than stopped of course

PR #526, opened 2024-03-21
operations send default values on construction

magni-mar:
#513


jbbjarnason (approved):
Would be nice to like reduce the duplicity

```cpp
auto constexpr print_error = [](std::string slot_name) { 
  return [slot_name](std::error_code const& err, std::size_t) 
    if (err) {
      logger_.info("...");
    }
 }
starting_.async_send(false, print_error("starting"));
// ...

```

PR #527, opened 2024-03-22
tfcctl for slots

magni-mar:


exes/tfcctl/src/tfcctl.cpp
```diff
@@ -76,6 +76,25 @@ inline auto stdin_coro(asio::io_context& ctx, std::string_view signal_name) -> a
...
+                      if (new_value.has_value()) {
+                        fmt::println("New value on slot: {}", new_value.value());
+                      } else {
+                        fmt::println("Error on slot: {}", new_value.error());
+                      }
+                    } else {
+                      fmt::println("New value on slot: {}", new_value);
+                    }
+                  } };
+  ctx.run();
```
omarhogni, line 95: This is not a good placement for running the context.

omarhogni (changes requested):

exes/tfcctl/src/tfcctl.cpp
magni-mar: the problem is the scope for the slot, i cannot declare the slot before deducing the type because of the constructor, declaring a variant also doesn't work.

magni-mar: next commit does have an alterior solution using optionals, take a look at that

magni-mar: The current solution is the only one that  I have found to work

omarhogni: You need to better structure the lifetimes expressed here. Lets look better into this before merging. You can still use this version for local testing.

jbbjarnason: > the problem is the scope for the slot, i cannot declare the slot before deducing the type because of the constructor, declaring a variant also doesn't work.

How does the variant fail? can this not be used: https://github.com/Skaginn3x/framework/blob/main/libs/ipc/inc/public/tfc/ipc.hpp#L216

jbbjarnason (commented):
Would be nice to know why the variant case fails, as mentioned in comment thread.

jbbjarnason (approved):

Issue #498, opened 2024-02-17
Invert filter needs to update value 


jbbjarnason:
When adding or removing a invert a event needs to be placed to let the logic handle the new "state"

magni-mar:
#528

PR #528, opened 2024-03-22
update value on slot when filter is added

magni-mar:
#498


jbbjarnason (commented):
I like how simple this is, but can we skip the dependence on matching dbus and use observable within filter.hpp and propagate the event from there only if changes to invert filter

magni-mar:
> I like how simple this is, but can we skip the dependence on matching dbus and use observable within filter.hpp and propagate the event from there only if changes to invert filter

I've tried now for quite some time to do that but to no success, the vector that owns the filters doesn't like to be changed into an observable. If you want to take a crack at this you can change this line in `filters.hpp` to an observable:

```cpp
confman::config<config_t> filters_;
```

```cpp
confman::config<tfc::confman::observable<config_t>> filters_;
```

or change it into a struct but that achieves the same result. I think the dependancy to sdbus matcher is the cleanest solution, as is in the first commit.

jbbjarnason:
> > I like how simple this is, but can we skip the dependence on matching dbus and use observable within filter.hpp and propagate the event from there only if changes to invert filter
> 
> I've tried now for quite some time to do that but to no success, the vector that owns the filters doesn't like to be changed into an observable. If you want to take a crack at this you can change this line in `filters.hpp` to an observable:
> 
> ```c++
> confman::config<config_t> filters_;
> ```
> 
> ```c++
> confman::config<tfc::confman::observable<config_t>> filters_;
> ```
> 
> or change it into a struct but that achieves the same result. I think the dependancy to sdbus matcher is the cleanest solution, as is in the first commit.

I have made a working solution with observable here, https://github.com/Skaginn3x/framework/pull/537 but need tests. Though it has been rigorously tested manually.

omarhogni:
Closing in favor of #545

PR #529, opened 2024-03-26
Added raw editing to configurator

SindriTh:
Implements a way to edit configuration objects as raw JSON with syntax highlighting.

image
image


omarhogni:
Looking good, would it be possible to add a backup/restore buttons?

SindriTh:
> Looking good, would it be possible to add a backup/restore buttons?

That should not be an issue. Would you like to store it as a JSON on the client? (Would be implemented as a default save/open file window)

SindriTh:
Updated screenshot to show new buttons @omarhogni

omarhogni (approved):

PR #531, opened 2024-04-04
Modify FAO marking, added from string converter and fixes from_int for strings beginning with A

omarhogni:


libs/ipc/testing/tests/database.csv
jbbjarnason, line 1: Is it not possible to change this to a source file and use static assert instead for each line?

omarhogni: I can add that to be part of the test compilation unit. There are allot of species and I would prefer not to run that every time someone uses item.hpp.

docs/functionality/fish_species_encoding.md
```diff
@@ -0,0 +1,100 @@
...
+indepent of the customer using our systems we have utilized
+a predefined list of fish species obtained from fao see https://www.fao.org/fishery/en/collection/asfis/en
+
+In addition to having a defined list of fish species we have a system
+of encoding the three letter markings into integers that makes it easy
+to transmit to plc controllers and any other system that requires knowledge
+of the product being processed.
+
+Due to our requirement of plc's having an easy time understanding the
+encoding we limited the size of this integer to 16-bits. This makes it
```
jbbjarnason, line 13:
```suggestion
encoding, we limited the size of this integer to 16-bits. This makes it
```

docs/functionality/fish_species_encoding.md
```diff
@@ -0,0 +1,100 @@
...
+
+```
+Here is an example for the fish species `HAD`
+```
+(26^2 * 7) + (26 * 0) + 3 = 473
+```
+
+## Speciality markings
+Anybody can see from this example and the fact that we limited the size of our integer to ```2^16```
+that there is still some space. We have created `metadata` that can also be sent as a specie.
```
jbbjarnason, line 63:
```suggestion
that there is still some space. We have created `metadata` that can also be sent as a species.
```

libs/ipc/inc/public/tfc/ipc/item.hpp
```diff
@@ -70,8 +70,29 @@ struct destination {
 namespace fao {
 struct species {
   bool outside_spec{ false };  // if struct is used for outside the specification, represented with `!` as first character
-  tfc::stx::basic_fixed_string<char, 3> code{};
-
+  using string_type = tfc::stx::basic_fixed_string<char, 3>;
+  string_type code{};
+  static constexpr auto from_3a(const std::string_view markedCode) -> std::optional<species> {
```
jbbjarnason, line 75:
```suggestion
  static constexpr auto from_3a(const std::string_view marked_code) -> std::optional<species> {
```

libs/ipc/inc/public/tfc/ipc/item.hpp
```diff
@@ -84,22 +105,26 @@ struct species {
       res.code[cnt] = alphabet[input % alphabet.size()];
       input /= alphabet.size();
     }
+    while (cnt > 0) {
+      --cnt;
+      res.code[cnt] = 'A';
+    }
```
jbbjarnason, line 111: maybe comment on which edge case this solves

libs/ipc/inc/public/tfc/ipc/item.hpp
```diff
@@ -113,10 +138,8 @@ struct species {
   }
   constexpr auto operator==(species const& rhs) const noexcept -> bool = default;
 
-  static constexpr std::string_view alphabet{ "ABCDEFGHIJKLMNOPQRSTUVWXYZ!" };  // note the ending !
-  // Todo remake labelled database, this offset is the result of a database labelling error, as in the - 1
-  // todo this can overlap !!!
-  static constexpr std::uint16_t offset{ (alphabet.size() - 1) * (alphabet.size() - 1) * (alphabet.size() - 1) };
+  static constexpr std::string_view alphabet{ "ABCDEFGHIJKLMNOPQRSTUVWXYZ" };  // note the ending !
```
jbbjarnason, line 141:
```suggestion
  static constexpr std::string_view alphabet{ "ABCDEFGHIJKLMNOPQRSTUVWXYZ" };
```

jbbjarnason (approved):

libs/ipc/testing/tests/database.csv
omarhogni: Done

libs/ipc/inc/public/tfc/ipc/item.hpp
```diff
@@ -84,22 +105,28 @@ struct species {
       res.code[cnt] = alphabet[input % alphabet.size()];
       input /= alphabet.size();
     }
+    // If the string remainder is only as the input value is 0, but
```
omarhogni, line 108:
```suggestion
    // If the string remainder is only a's the input value is 0, but
```

PR #546, opened 2024-04-17
add length, pressure, temperature to ipc

jbbjarnason:


omarhogni (approved):
Can we try to have the temperature unit in Celsius?

jbbjarnason:
> Can we try to have the temperature unit in Celsius?

Now as µ˚C

PR #553, opened 2024-05-14
add dimension

magni-mar:


libs/stx/inc/public/tfc/utils/units_glaze_meta.hpp
```diff
@@ -178,6 +178,8 @@ inline consteval auto dimension_name() -> std::string_view {
     return "pressure";
   } else if constexpr (mp_units::convertible(ref_t, mp_units::si::kelvin)) {
     return "temperature";
+  } else if constexpr (mp_units::convertible(ref_t, mp_units::si::litre / mp_units::si::second)) {
+    return "flow";
```
omarhogni, line 182: Could you call this `volumetric_flow`  and remove the comment above with volumetric_flow L#120.

omarhogni (changes requested):

libs/stx/inc/public/tfc/utils/units_glaze_meta.hpp
magni-mar: finished

omarhogni (approved):

PR #554, opened 2024-05-14
add minutes to glaze

magni-mar:


jbbjarnason:
@magni-mar  Could you provide an example of where this is used/needed?

magni-mar:
> @magni-mar Could you provide an example of where this is used/needed?

When a config uses `std::chrono::minutes`

libs/stx/inc/public/tfc/stx/glaze_meta.hpp
```diff
@@ -53,6 +53,8 @@ inline constexpr auto make_ratio_symbol() -> std::string_view {
     return "P";
   } else if constexpr (std::is_same_v<type, std::exa>) {
     return "E";
+  } else if constexpr (std::is_same_v<type, std::ratio<60>>) {
+    return "minute";
```
jbbjarnason, line 57:
```suggestion
    return "ratio<60>";
```

jbbjarnason (approved):

PR #556, opened 2024-05-15
add filter for pressure

magni-mar:


jbbjarnason (approved):
Good catch! I will follow up with the other missing types.

PR #559, opened 2024-05-16
update toolchain to gcc-14

jbbjarnason:
Resolved minor issues with compiling.

This is now using my vcpkg fork, awaiting update of glaze: https://github.com/microsoft/vcpkg/pull/38819

### DIFF
--- a/containers/tfc-toolchain/install-common.sh
+++ b/containers/tfc-toolchain/install-common.sh
@@ -29,7 +29,6 @@ hide_output apt install -y --no-install-recommends \
 
 # Install documentation specific packages
 hide_output apt install -y --no-install-recommends \
-  doxygen \
   python3-venv \
   python3-jinja2 \
   graphviz
--- a/containers/tfc-toolchain/install-common.sh
+++ b/containers/tfc-toolchain/install-common.sh
@@ -20,7 +20,6 @@ hide_output apt install -y --no-install-recommends \
   libtool \
   autopoint \
   gperf \
-  python3-distutils \
   python3-jinja2 \
   python3-venv \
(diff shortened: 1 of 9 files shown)


containers/tfc-toolchain/install-common.sh
```diff
@@ -20,17 +20,15 @@ hide_output apt install -y --no-install-recommends \
...
-  g++-13 \
-  g++-13-aarch64-linux-gnu \
+  gcc-14 \
+  g++-14 \
+  g++-14-aarch64-linux-gnu \
   libltdl-dev # Required for autoconf
 
 # Install documentation specific packages
 hide_output apt install -y --no-install-recommends \
-  doxygen \
```
jbbjarnason, line 33: Now built from source, the doxygen binary had some strange llvm runtime issues, probably because of two differently build llvm libraries in the container, from apt and our built from source

libs/stx/inc/public/tfc/stx/glaze_meta.hpp
```diff
@@ -194,12 +194,12 @@ template <typename duration_t>
 constexpr auto parse8601(const std::string& save) -> date::sys_time<duration_t> {
   std::istringstream in{ save };
   date::sys_time<duration_t> tp;
-  in >> date::parse("%FT%TZ", tp);
```
jbbjarnason, line 197: This resolved in compilation errors, ambiguity, I was not able to see through how and why it only occurs with gcc14, so now using explicit function from_stream

jbbjarnason:
@omarhogni I will merge this PR, please raise an issue if you would like something changed.

PR #563, opened 2024-05-21
remove namespace decl

magni-mar:


jbbjarnason (approved):
I don't see the necessity of those changes but sure they work.

PR #587, opened 2024-06-12
Inital draft

omarhogni:


jbbjarnason (commented):
Two discussion topics:

- Where would we like to convert between Meters and Feet etc.
- Default values for a fmt::arg

PR #593, opened 2024-06-25
Update all dep

jbbjarnason:


libs/ipc/inc/public/tfc/ipc/details/dbus_server_iface.hpp
omarhogni, line 1: Should we rather throw here?

libs/ipc/inc/public/tfc/ipc/details/dbus_ipc.hpp
omarhogni, line 1: Could be better to throw here also.

exes/mqtt-bridge/inc/client.hpp
```diff
@@ -173,13 +171,13 @@ class client {
...
-      for (uint64_t i = 0; i < publish_packet->payload().size(); i++) {
-        process_payload(publish_packet->payload()[i], *publish_packet);
+      for (auto const& itm : publish_packet->payload_as_buffer()) {
+        process_payload(itm, *publish_packet);
       }
     }
   }
 
-  auto strand() -> asio::strand<asio::any_io_executor> { return endpoint_client_->strand(); }
+  auto get_executor() -> asio::any_io_executor { return endpoint_client_->get_executor(); }
```
omarhogni: const

omarhogni (approved):

